PR #2, opened 2016-03-20
Converting tabs to spaces and renaming dependency to avoid conflicts

tarciosaraiva:
`check-types` requirement could conflict with the `check()` function in the code base. Fixed that and converted all tabs to spaces.


mboudreau:
I've added these changes in manually while changing a few things around in the library, incrementing the version number to 2.0.0 since these changes can be breaking.

PR #3, opened 2016-05-15
Pact provider verification feature

mefellows:
* Added the ability to run a Pact Verification process (uses a standalone Ruby Gem: https://github.com/pact-foundation/pact-provider-verifier)
* Verifies multiple local/remote Pacts in one go
* Added Provider integration tests to demonstrate usage
* Added Pact Broker authentication
* Updated docu
* Refactored library so that the Mock Service did not take precedence over all APIs. 

*NOTE*: This will need a major version bump due to the API change (x() => xServer() all the things). Further note, this cannot be merged until the https://github.com/pact-foundation/pact-provider-verifier-npm Travis deployment succeeds.


mefellows:
https://travis-ci.org/pact-foundation/pact-provider-verifier-npm now green - I think we are good to merge and update version!

src/logger.js
```diff
@@ -8,7 +8,7 @@ var config = {
 	name: 'pact-node@' + pkg.version,
 	streams: [
 		{
-			level: 'info',
+			level: 'debug',
```
mboudreau: Make this info again.  Maybe use an environment variable to flick between dev and release version?

src/integration/provider.js
```diff
@@ -0,0 +1,102 @@
+var cors = require('cors'),
```
mboudreau: Move this file under 'test' directory

scripts/check-dependencies.js
```diff
@@ -0,0 +1,17 @@
+var pkg = require('../package.json');
```
mboudreau: I had removed this in a past commit since I moved it to pact-mock-service as a wrapper.  Should probably do the same with pact-provider-verifier.

mefellows: Annoying little script keeps coming back!

mefellows:
Have implemented that feedback Michel - will do the same for the publish task too (#4).

PR #6, opened 2016-05-29
Optional logging

tarciosaraiva:
A couple of improvements. I've used `pact-node` on another project and prefer it to be less verbose. So here's a PR along with a couple of other enhancements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Simply require the library and call the create function.
 
 ```js
 var pact = require('@pact-foundation/pact-node');
-var server = pact.create({port: 9999});
+var server = pact.createServer({port: 9999});
 server.start().then(function() {
 	// Do your testing/development here
 });
--- a/package.json
+++ b/package.json
@@ -82,10 +82,11 @@
   "scripts": {
     "clean": "rimraf logs/*",
     "lint": "jscs .",
+    "pretest": "npm run clean && npm run lint",
     "test": "./node_modules/.bin/cross-env DEVELOPMENT=true ENABLE_LOG=true mocha ./**/*.spec.js",
     "watch": "nodemon -x npm run dev",
     "watch:debug": "nodemon --debug -q -w assets/ --ext '.' --exec 'npm run lint'",
-    "dev": "npm run clean && npm run lint --force && npm test && ./node_modules/.bin/cross-env DEVELOPMENT=true node .",
+    "dev": "npm test && ./node_modules/.bin/cross-env DEVELOPMENT=true node .",
     "start": "npm run watch"
   },
   "homepage": "https://github.com/pact-foundation/pact-node#readme"
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
   "scripts": {
     "clean": "rimraf logs/*",
     "lint": "jscs .",
-    "test": "./node_modules/.bin/cross-env DEVELOPMENT=true mocha -R spec ./**/*.spec.js --timeout 15000",
+    "test": "./node_modules/.bin/cross-env DEVELOPMENT=true ENABLE_LOG=true mocha ./**/*.spec.js",
     "watch": "nodemon -x npm run dev",
     "watch:debug": "nodemon --debug -q -w assets/ --ext '.' --exec 'npm run lint'",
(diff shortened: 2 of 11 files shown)


mefellows:
I like it. LGTM.

mboudreau:
@tarciosaraiva can you please explain why you need to disable logging using environment variables? Why not set the log level to be higher?

mefellows:
I suppose there are cases that you don't want _any_ logs - so setting it higher won't help. But it might make sense for it to be enabled by default, and explicitly check for it to be _disabled_.

tarciosaraiva:
@mefellows LGTM?

@mboudreau Regarding environment variables: http://12factor.net/config
Setting the level higher won't achieve what I want, which is disable the logging completely. I saw that `logger.level()` is exposed and tried using that but didn't work for me. So I decided to send this PR.

Considering what @mefellows said I'm happy to invert the selection.
More over, I could check for the environment variable `NODE_ENV === 'test'` and stream that to a file.

What do you prefer?

mefellows:
Lgtm = looks good to me :)

Good point. Personally I'd probably want to control most things via env vars. I think i favour more explicit log redirection rather than adjust on the node env 'test' etc. But I get that is fairly typical in NodeJS

tarciosaraiva:
@mefellows well, it's a combination of both worlds but I don't really like that approach to be honest: feels more like a work around to preserve logs.

I prefer using env variables that have meaning... `DISABLE_LOGGING` is very clear.

mefellows:
We're on the same page - that's exactly the sort of variable I'm talking about. When in front of a PC I'll add more detail.

mefellows:
So basically, for logging to work, a client should be able to control the following:

* If logging should be enabled or not
* Where logs should be output (display/file/...)
* What level of logs should be logged to the output

I'm not suggesting we create environment variables for them all - if there is a code level API available (as we currently have for all but on/off), a client can choose to use environment variables in their code to call those APIs anyway. e.g. `logger.level(process.env.LOG_LEVEL)` type thing. 

Also, just in terms of disabling logging, the Bunyan author [suggests](https://github.com/trentm/node-bunyan/pull/148#issuecomment-63018668) that you do something like `log.level(bunyan.FATAL + 1)` do disable logs. Might be worth adding a method `disable()` or something. I think this is probably preferred to environment variables as this way there is a code-level API to control all the things, and then the client/user is free to implement around that as they choose ("12-factor"-style or otherwise).

Just my 2c

mboudreau:
Environment Variables is not an appropriate solution for a library.  We'll hit the same issues with it as RUBYGEMS_DEPS.  It won't be easy to document, or setup, and frankly, if a library I was using forces me to use env vars to setup functionality that isn't available through the API, I'd be pretty peeved.

I'm looking into solutions.  One could be to create an abstract interface (kinda the same way @tarciosaraiva  did within logger file) with a way to disable logging (which in this case means just removing the stream), or maybe just expose the bunyan logger instance through the API.  I'm still researching a few options.

mefellows:
Could we achieve that with another [export](https://github.com/pact-foundation/pact-node/pull/6/files#diff-567859a4cc8b910de0454da79959bde6R25) that lets us enable/disable the logs and also set the output [stream](https://github.com/trentm/node-bunyan#streams)?

e.g.
```
disable: function () {
   log.level(bunyan.FATAL + 1);
},
setStream: function (stream) {
   // https://github.com/trentm/node-bunyan#streams
   ...
},
```

tarciosaraiva:
@mboudreau you're not really using environment variables to setup functionality. If you don't use, it will behave exactly the same way as before (assuming that I use the DISABLE_LOGGING). Said that though I get your point.

Using the `disable` as suggested by @mefellows is really ideal but I had a few issues with that, related to an exception leading to streams not available. I wonder if the `.bind`s you have in the test code are having some sort of side effect in there.

Going back to @mefellows suggestion they also have `bunyan.mute = true;` - might try that.

mboudreau:
tarcio, yet, there's already some good functionality in bunyan.  I'll merge your stuff for now because I think you did fix a few things along the way, but I will change them tonight to reflect what we've spoken about.

Issue #7, opened 2016-06-08
Verifier doesn't return error information to client


mefellows:
[Currently](https://github.com/pact-foundation/pact-node/blob/master/src/verifier.js#L90-L107) we only log Verifier output (using Bunyan), it would be preferable to have the output returned in the promise chain so that we can properly handle it in downstream libs (pact-js, mocha etc.). The current message 'Pact Verification failed' is insufficient information, and too late to make a programmatic call to output logs.

When things succeed, this data is generally not required and is noise, but when things fail we need to be able to programatically access that information.

I suggest we pipe the stderr/out to a variable and return it in the promise rejection (and success) at https://github.com/pact-foundation/pact-node/blob/master/src/verifier.js#L104.

mboudreau:
@mefellows is this good to be closed?

mefellows:
Yep!

PR #8, opened 2016-06-08
Return output to promise in verification process

mefellows:
Fixes issue #7.

Essentially, we log all output as before, but we also store it in a string variable that is passed back to the promise chain. This allows clients (such as pact-foundation/pact-js) to provide better messaging to their users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pact-foundation/pact-node",
-  "version": "4.2.0",
+  "version": "4.2.1",
   "description": "A wrapper for the Ruby version of Pact to work within Node",
   "main": "./src/pact.js",
   "os": [


mboudreau:
@mefellows Check why the build is failing and please use the same promise library throughout the project.  We can't have a mismash of libraries trying to do the same thing.

mefellows:
Not sure what you mean with promise libs, I didn't change that (AFAIK). Weirdly the build is failing intermittently - never a good thing.

mboudreau:
You added Promise.reject within an integration test, which doesn't resolve to anything:

```
return expect(verifier.verify().catch(function(err) {
							if (!_.isEmpty(err) && err.indexOf("unexpected token at 'Unauthorized'") > -1) {
								return Promise.reject(new Error("failed"));
							}
						})).to.eventually.be.rejectedWith(Error, "failed");
```

You should be using q.reject instead. I'm guessing the build is failing intermittently because of an async test that calls done too quickly.

mefellows:
Right. I didn't add a new Promise in this PR though (look at the diff) so
that's a bit weird. I'll look into it, thanks for digging through it.

On Thu, Jun 9, 2016 at 11:05 AM, Michel Boudreau <notifications@github.com>
wrote:

> You added Promise.reject within an integration test, which doesn't resolve
> to anything:
>
> return expect(verifier.verify().catch(function(err) {
>                             if (!_.isEmpty(err) && err.indexOf("unexpected token at 'Unauthorized'") > -1) {
>                                 return Promise.reject(new Error("failed"));
>                             }
>                         })).to.eventually.be.rejectedWith(Error, "failed");
>
> You should be using q.reject instead. I'm guessing the build is failing
> intermittently because of an async test that calls done too quickly.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/pact-foundation/pact-node/pull/8#issuecomment-224775840>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AADSjC1Q91s8l5580T8RyuyZJrSXUeUZks5qJ2bKgaJpZM4IxbIH>
> .
>



-- 
Matt Fellows
- http://www.onegeek.com.au/

PR #10, opened 2016-06-19
Validate remote Pact files on publish

mefellows:
Addresses comment: https://github.com/pact-foundation/pact-node/commit/a5afdb17e2c3ca8acf6c6a2051e7575cc4bcb4b2?diff=unified#commitcomment-17854867

New checks:
* Existence of remote files (will use basic auth if provided)
* Sanity check that a consumer/provider has been added to the Pact file

### DIFF
--- a/src/publisher.js
+++ b/src/publisher.js
@@ -28,10 +28,10 @@ Publisher.prototype.publish = function () {
 	var uris = _.chain(options.pactUrls)
 		.map(function (uri) {
 			var localFileOrDir = path.normalize(uri);
-			if (!(/^http/.test(uri)) && fs.statSync(localFileOrDir).isDirectory()) {
+			if (!(/^http/.test(uri)) && fs.statSync(localFileOrDir) && fs.statSync(localFileOrDir).isDirectory()) {
 				uri = localFileOrDir;
 				return _.map(fs.readdirSync(uri, ''), function (file) {
-					if (/.json$/.test(file)) {
+					if (/\.json$/.test(file)) {
 						return path.join(uri, file);
 					}
 				});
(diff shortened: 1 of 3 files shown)


src/publisher.js
```diff
@@ -27,11 +27,10 @@ Publisher.prototype.publish = function () {
 	// publish template $pactHost/pacts/provider/$provider/consumer/$client/$version
 	var uris = _.chain(options.pactUrls)
 		.map(function (uri) {
-			uri = path.normalize(uri);
-			if (fs.statSync(uri).isDirectory()) {
-				logger.debug('we are a dir: ' + uri);
+			var localFileOrDir = path.normalize(uri);
+			if (!(/^http/.test(uri)) && fs.statSync(localFileOrDir).isDirectory()) {
```
mboudreau: Doesn't fstat return undefined if the part is no good? This would cause an error of it tried to call isDirectory.

src/publisher.js
```diff
@@ -46,42 +45,83 @@ Publisher.prototype.publish = function () {
...
-					user: options.pactBrokerUsername,
-					pass: options.pactBrokerPassword
-				}
+			var deferred = q.defer();
+
+			// Promise to update provider/consumer
+			var getPactCollaborators;
+
+			// Parse the Pact file to extract consumer/provider names
+			if (/.json$/.test(uri)) {
```
mboudreau: You need to escape the dot in regex

src/publisher.js
mefellows: Probably. Although that was the case prior to this PR also.

src/publisher.js
mefellows: Indeed we do. Similarly, that was the case prior to this PR also.

src/publisher.js
mboudreau: Yeah, but doesn't mean we can't fix it :P

mefellows:
Fixed the regexes missing escape (i.e. `\.json`) and also the potential call on an `undefined`. Good spots and worth tidying up.

Issue #12, opened 2016-06-21
Publisher issues


tarciosaraiva:
I pulled the code to try and submit a fix but the tests keep failing - even on a clean clone - so I decided to write an issue here.

#### Issue 1
Under `publisher.js` on line 99 it uses `urlJoin` to create a URL to PUT to pact-broker. The issue here is related to `data.provider` and `data.consumer` which are both objects. You have to retrieve the actual name by using `data.provider.name` and `data.consumer.name` otherwise the URL built will look like

```
/pacts/provider/[Object%20object]/consumer/[Object%20object]/version/2.0
```

#### Issue 2
When there's a failure there's no way to know what happened. On `publisher.js` line 114 pass back the response object when the promise is rejected.

Thanks

mboudreau:
@mefellows is this related to the change you did/deployed this morning?

mefellows:
Looks like it, but not 100% (off the top of my head). There is a unit test associated with it but it's clearly insufficient. I'll investigate this evening if I can.

tarciosaraiva:
Sorry to be a pain but there's only issue addressed on that commit

mefellows:
Issue 1, do you mean?

tarciosaraiva:
Yup

mefellows:
All good, WIP mate :)

tarciosaraiva:
@mefellows @mboudreau somehow version 4.4.3 didn't get published. Can someone act on it? I have a project here needing this change.

mboudreau:
@tarciosaraiva The fix branch is building.  Waiting to go green before merging and deploying the latest.  Keep your pants on ;)

tarciosaraiva:
I think this is good now. Cheers

mefellows:
👍

PR #13, opened 2016-06-22
Validate options when generating pact publish URL

mefellows:
Fixes issue #12:

* Returns a rejected promise with a proper error message
* Validates a that a broker url is provided and a valid consumer/provider object exists in the Pact file


mefellows:
Travis CI issues seem to have disappeared and build is green again (and running nice and quick). I've currently left Mac OSX builds out - do we want them back? Not sure how much value they're adding.

mboudreau:
Please do.  I want to make sure all our code works in all OS'.  We've been
burned by this before and I don't want it happening again.

mefellows:
Np, I have added them back in. Any objections to merging this now (Mac builds still seem to be affected in TravisCI)? Normally I would wait, but this PR fixes a problem we have released so I don't see harm in releasing on the small chance it doesn't build on OSX (and even then I'll be skeptical as I build on OSX locally).

mefellows:
OK, we're proper green now. Err, well sorta (looking at you Appveyor!).

PR #14, opened 2016-06-29
adding tagging capability

tarciosaraiva:
the pact broker allows for an endpoint that takes only one tag as specified here: https://github.com/bethesque/pact_broker/wiki/How-to-ensure-backwards-compatibility-by-tagging-pacts

This change creates an array of promises to loop through the array of tags and sends a PUT request to the Pact Broker accordingly, returning the original `deferred` object when all is done.

Feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,8 @@ var opts = {
 	pactUrls: <Array>,               // Array of local Pact files or directories containing them. Required.
 	pactBroker: <String>,            // URL to fetch the provider states for the given provider API. Optional.
 	pactBrokerUsername: <String>,    // Username for Pact Broker basic authentication. Optional
-	pactBrokerPassword: <String>     // Password for Pact Broker basic authentication. Optional
+	pactBrokerPassword: <String>     // Password for Pact Broker basic authentication. Optional,
+  tags: <Array>,                   // An array of Strings to tag the Pacts being published. Optional
 };
 
 pact.publishPacts(opts)).then(function () {
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ var opts = {
 	pactBroker: <String>,            // URL to fetch the provider states for the given provider API. Optional.
 	pactBrokerUsername: <String>,    // Username for Pact Broker basic authentication. Optional
 	pactBrokerPassword: <String>     // Password for Pact Broker basic authentication. Optional,
-  tags: <Array>,                   // An array of Strings to tag the Pacts being published. Optional
+	tags: <Array>,                   // An array of Strings to tag the Pacts being published. Optional
 };
 
 pact.publishPacts(opts)).then(function () {
--- a/src/publisher.js
+++ b/src/publisher.js
@@ -27,6 +27,10 @@ var constructPutUrl = function(options, data) {
 		throw new Error("Cannot construct Pact publish URL: 'pactBroker' not specified");
 	}
 
+	if (!_.has(options, 'consumerVersion')) {
+		throw new Error("Cannot construct Pact publish URL: 'consumerVersion' not specified");
+	}
+
 	if (!_.isObject(options)
 		|| !_.has(data, 'consumer')
 		|| !_.has(data, 'provider')
@@ -41,7 +45,11 @@ var constructPutUrl = function(options, data) {
 
 var constructTagUrl = function(options, tag, data) {
 	if (!_.has(options, 'pactBroker')) {
-		throw new Error("Cannot construct Pact publish URL: 'pactBroker' not specified");
+		throw new Error("Cannot construct Pact Tag URL: 'pactBroker' not specified");
+	}
+
+	if (!_.has(options, 'consumerVersion')) {
+		throw new Error("Cannot construct Pact Tag URL: 'consumerVersion' not specified");
 	}
 
 	if (!_.isObject(options)
--- a/src/publisher.js
+++ b/src/publisher.js
@@ -60,12 +60,8 @@ Publisher.prototype.publish = function () {
 							'Accept': 'application/json'
 						},
 						body: data
-					}).then(function (response) {
-						if (response.statusCode >= 200 && response.statusCode < 300) {
-							return data;
-						} else {
-							return q.reject(new Error('Unable to publish Pact to Broker: ' + response.statusCode));
-						}
+					}).fail(function (err) {
+						return q.reject(new Error('Unable to publish Pact to Broker. ' + err.message));
 					});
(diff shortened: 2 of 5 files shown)


mefellows:
Another random failure on Node 4.0 on OSX. Based on previous PRs, I'll just kick it in the guts and see if it resolves it...

tarciosaraiva:
I suppose this comment was to be on the other PR 😉

mefellows:
That is _weird_... woops!

README.md
```diff
@@ -80,7 +80,8 @@ var opts = {
 	pactUrls: <Array>,               // Array of local Pact files or directories containing them. Required.
 	pactBroker: <String>,            // URL to fetch the provider states for the given provider API. Optional.
 	pactBrokerUsername: <String>,    // Username for Pact Broker basic authentication. Optional
-	pactBrokerPassword: <String>     // Password for Pact Broker basic authentication. Optional
+	pactBrokerPassword: <String>     // Password for Pact Broker basic authentication. Optional,
+  tags: <Array>,                   // An array of Strings to tag the Pacts being published. Optional
```
mefellows: Better fix this one.

src/publisher.js
```diff
@@ -38,6 +39,21 @@ var constructPutUrl = function(options, data) {
 	return urlJoin(options.pactBroker, 'pacts/provider', data.provider.name, 'consumer', data.consumer.name, 'version', options.consumerVersion)
 };
 
+var constructTagUrl = function(options, tag, data) {
```
mefellows: We probably ought to have a check for `consumerVersion` here too.

src/publisher.js
```diff
@@ -130,21 +147,59 @@ Publisher.prototype.publish = function () {
 
 					http(config, function (error, response) {
 						if (!error && response.statusCode == 200) {
```
mefellows: This might end up being a conflict with #16, but should be easy enough to resolve.

mefellows:
LGTM aside from a couple of very minor things - great stuff man!

As a general whinge, it is clear (to me, at least) that superagent+bluebird would have looked a hell of a lot prettier than the code we ended up with here now that we have to make a few more HTTP requests (much more concise/flat). There is utility in reducing dependencies, so 6 in one, half a dozen the other I suppose.

tarciosaraiva:
@mefellows thanks for taking the time to have a look at it. Yes, superagent would be much nicer. That's another discussion though, I'm happy to have a look at it later on.

mefellows:
Yep, let's not change it for now.

On Thu, Jun 30, 2016 at 7:33 AM, Tarcio Saraiva <notifications@github.com>
wrote:

> @mefellows <https://github.com/mefellows> thanks for taking the time to
> have a look at it. Yes, superagent would be much nicer. That's another
> discussion though, I'm happy to have a look at it later on.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/pact-foundation/pact-node/pull/14#issuecomment-229495787>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AADSjN91G6W9MvKB-XtXf6rMFken4_Krks5qQuScgaJpZM4JAub9>
> .
>



-- 
Matt Fellows
- http://www.onegeek.com.au/

src/publisher.js
```diff
@@ -130,21 +155,59 @@ Publisher.prototype.publish = function () {
...
 						if (!error && (response.statusCode >= 200 && response.statusCode < 300)) {
-							deferred.resolve();
+							publishDeferred.resolve(data);
 						} else {
 							if (error != null) {
-								deferred.reject(error);
+								publishDeferred.reject(error);
 							} else {
-								deferred.reject(new Error('Unable to publish Pact to Broker: ' + response.statusCode));
+								publishDeferred.reject(new Error('Unable to publish Pact to Broker: ' + response.statusCode));
```
mboudreau: Can't say I'm a fan of this conditional.  Seems overly complicated and can be simplified.

src/publisher.js
```diff
@@ -130,21 +155,59 @@ Publisher.prototype.publish = function () {
...
 					});
 
-					return deferred.promise;
+					return publishDeferred.promise;
 				}, function(err) {
 					return q.reject(err);
 				})
+				.then(function(data) {
+					if (!options.tags.length) {
+						deferred.resolve();
```
mboudreau: Where is this deferred coming from?!

src/publisher.js
```diff
@@ -130,21 +155,59 @@ Publisher.prototype.publish = function () {
...
+						});
+
+						return tagDeferred.promise;
+					}))
+				}, function(err) {
+					return q.reject(err);
+				})
+				.then(function() {
+					deferred.resolve();
+					return deferred.promise;
```
mboudreau: Why is this needed?

src/publisher.js
mefellows: Presumably line [102](https://github.com/pact-foundation/pact-node/pull/14/files/acc1c9a02e9bc1009fab6d2bcdf284530636a627#diff-b3f6177d1f5ad151f358f56f306500c1R102)

src/publisher.js
mefellows: There must be a way to flatten all of these promises. I might be a Bluebird fanboy (for no good reason other than I've used it(, but I'd find it hard to believe q can't do it too. No doubt it would have influenced Bluebirds design. I feel this would make all of this easier to read (to be fair, previous to this PR it wasn't pretty either and I'm probably at least in part to blame).

src/publisher.js
mefellows: From what I can tell it's to be able to return a more useful error message. Would you prefer something like:

```js
if (!error && (response.statusCode >= 200 && response.statusCode < 300)) {
  publishDeferred.resolve(data);
else if (error != null) {
  publishDeferred.reject(error);
} else {
  publishDeferred.reject(new Error('Unable to publish Pact to Broker: ' + response.statusCode));
}
```

src/publisher.js
mboudreau: Yeah, you can with Q.denodeify for the http call and then chaining the promises properly.

src/publisher.js
mboudreau: see, I'd prefer this:
```
if(error || response.statusCode < 200 || response.statusCode >= 300) {
  return publishDeferred.reject(error ? error : new Error('Unable to publish Pact to Broker: ' + response.statusCode));
}
return publishDeferred.resolve(data);
```

PR #15, opened 2016-06-29
reducing matrix according to Node LTS

tarciosaraiva:
So, according to node's own LTS release (https://github.com/nodejs/LTS) I think it's worth a discussion to reduce the number of Node environments the build runs on.

- Version 0.12 is in Maintenance status and its LTS status will end at the end of the year
- Version 4 is the Active LTS and will end in 2018
- Version 5 is a stable release very much like v4 but contains breaking changes that may affect big orgs
- Version 6 will be the Active LTS starting October this year and incorporates v5

In order to reduce build time, specially on Appveyor, I would like to suggest to limit ourselves to v0.12, v4 and v6 (all latest of them).

Unless there's a reliance on specific Node features from the releases in the current matrix, I think we can be pretty safe in assuming that the releases mentioned above will suit well whatever level of Node the community is focusing on.

I have been using `pact-node` with v0.12 and v4 and v6 and all of them have been pretty solid. I'm not sure if Node v5 really adds anything however, if you want to read more, I suggest you read this: https://nodejs.org/en/blog/community/node-v5/

Happy to discuss.


mefellows:
I see your point with the 5.x branch - it's already outdated and not designed for 'enterprises'. If it breaks, we'd recommend moving to 6/7 which would be the official Node position also. Personally, with Appveyor though I'd be inclined to drop to a single version.

The fact that this was pushed 9 hours ago and we still don't have feedback from a build is telling. 

Similarly we're having issues with OSX on Travis. Almost all of my PRs in the past month have resulted in my manually restarting of sub-jobs where things magically work again. i.e. https://travis-ci.org/pact-foundation/pact-node/jobs/141008463 I've kicked off twice and the agent just never starts.

Linux/Open Source 1, Windows/OSX/Closed source 0

tarciosaraiva:
@mefellows Yup Appveyor is a special pain. Not sure why it takes so long and even now, more than 12 hours after the PR, the whole thing still haven't triggered. Perhaps we should look for an alternative?

mefellows:
Happy to if there is something good, but to be honest, setting it up was a certain kind of pain no-one wants to have to go through more than once - it would need to be good :). The best I can think of is BuildKite but we would need to manage our own infra which I'm not sure we want to do.

tarciosaraiva:
Ha ya, been there. We don't want that.

mboudreau:
I'm happy to reduce the matrix for appveyor, like 0.12, 4.x (latest) and 6.x latest

tarciosaraiva:
@mboudreau Cool. What about the Travis matrix? Yesterday 2 hours went by before the first notification email got out. Travis is building the same thing 18 times. Is there a motivation?

mboudreau:
@tarciosaraiva can't say I've ever seen travis take 2 hours to send an email, but yeah, we could reduce it down to 0.12, latest 4.x, latest 5.x and latest 6.x, but needs to be for all OS.  This is important since we have been burned in the past about a breaking change specific to an OS.

tarciosaraiva:
Excellent, thanks @mboudreau - I will push a change right away!

mboudreau:
@tarciosaraiva awesome, thanks.

mboudreau:
looks good, but you could remove node 5 from appveyor build since 6 has the same features.  If it works in 6, it should work for 5.  That's if you want to make appveyor as small as possible.

tarciosaraiva:
Excellent, done @mboudreau !

mboudreau:
Perfect.  looks good to me, merging.

PR #18, opened 2016-07-03
add an '.editorconfig' file for consistent code style across editors

tarciosaraiva:
Basically this: http://editorconfig.org/

Install the (plugin to your favourite editor)[http://editorconfig.org/#download] and watch it magically transform your code. It works pretty well.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


mefellows:
👏  good idea

mboudreau:
Yep. Agreed.

PR #19, opened 2016-07-07
fix(tag): fix URL for Pact tagging

tarciosaraiva:
A couple of things:
First, the URL was wrong, oversight on my part as it should have been `../versions/..` instead of `../version/..`

Then the `json: true` configuration on the module `request` updates the headers with Accept and Content-Type pointing to JSON. Unfortunately the Pact Broker was rejecting this with a HTTP 406...

> 406 Not acceptable
> The resource is valid, but cannot be provided in a format specified in the Accept headers in the request.

Fixed that too.


mboudreau:
Looks good.

PR #25, opened 2016-10-04
fix(cmd): quote command line invocation arguments, fixes #22

### LINKED ISSUE #22: having trouble verifying pacts from broker
when running `verifyPacts` with `pactUrls` pointing to local files, it is successfully doing verification of the pact. however, as soon as i try to change the url to point to a pact on my remote broker (https and using basic auth), i get the following error:

```
[Error: /bin/sh: <part of the password>: command not found
/bin/sh: <a different part of the password>: command not found
/Users/admin/development/travi.org/api/node_modules/@pact-foundation/pact-provider-verifier-darwin/lib/app/lib/pact/provider_verifier/app.rb:32:in `get_pact_consumer_name': undefined method `[]' for nil:NilClass (NoMethodError)
	from /Users/admin/development/travi.org/api/node_modules/@pact-foundation/pact-provider-verifier-darwin/lib/app/lib/pact/provider_verifier/app.rb:55:in `block in verify_pacts'
	from /Users/admin/development/travi.org/api/node_modules/@pact-foundation/pact-provider-verifier-darwin/lib/app/lib/pact/provider_verifier/app.rb:54:in `collect'
	from /Users/admin/development/travi.org/api/node_modules/@pact-foundation/pact-provider-verifier-darwin/lib/app/lib/pact/provider_verifier/app.rb:54:in `verify_pacts'
	from /Users/admin/development/travi.org/api/node_modules/@pact-foundation/pact-provider-verifier-darwin/lib/app/lib/pact/provider_verifier/cli.rb:18:in `verify'
	from /Users/admin/development/travi.org/api/node_modules/@pact-foundation/pact-provider-verifier-darwin/lib/vendor/ruby/2.1.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
	from /Users/admin/development/travi.org/api/node_modules/@pact-foundation/pact-provider-verifier-darwin/lib/vendor/ruby/2.1.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
	from /Users/admin/development/travi.org/api/node_modules/@pact-foundation/pact-provider-verifier-darwin/lib/vendor/ruby/2.1.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
	from /Users/admin/development/travi.org/api/node_modules/@pact-foundation/pact-provider-verifier-darwin/lib/vendor/ruby/2.1.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
	from /Users/admin/development/travi.org/api/node_modules/@pact-foundation/pact-provider-verifier-darwin/lib/app/pact-provider-verifier.rb:3:in `<main>'
]
```

note that the pact that i was using locally was a file fetched directly from the same url and credentials using postman.

i would expect that if it were failing to fetch from the broker, that i would get an error related to the fetch, but i'm not sure how to get more information about the context of the error. i did notice that parts of my password for the broker show up in the error (the minor edits in the output above). they seem to be split after `&` characters. if this is the case, i'm surprised that i don't get a "failed to auth" error before it tries to parse for a consumer name.

could you provide some guidance about how to dig deeper? is there a way to put the ruby portion into a debug mode? if this issue is that special characters need to be escaped, i'm not sure what the sequence would need to be here since i am passing the string as an option to `verifyPacts`. could it be updated to do the necessary escaping automatically?

mefellows:
Wraps all arguments to the verifier in quotation marks, to allow for arguments with special characters (see #22 for example with passwords).

### DIFF
--- a/src/verifier.js
+++ b/src/verifier.js
@@ -21,25 +21,27 @@ function Verifier(providerBaseUrl, pactUrls, providerStatesUrl, providerStatesSe
 	this.options.providerStatesSetupUrl = providerStatesSetupUrl;
 	this.options.pactBrokerUsername = pactBrokerUsername;
 	this.options.pactBrokerPassword = pactBrokerPassword;
+	this.options.args = [];
 }
 
 Verifier.prototype.verify = function () {
 	logger.info("Verifier verify()");
 	var deferred = q.defer();
-	
+
 	var output = ''; // Store output here in case of error
 	function outputHandler(data) {
 		logger.info(data);
 		output += data;
 	}
-	
+
 	var envVars = JSON.parse(JSON.stringify(process.env)); // Create copy of environment variables
 	// Remove environment variable if there
 	// This is a hack to prevent some weird Travelling Ruby behaviour with Gems
 	// https://github.com/pact-foundation/pact-mock-service-npm/issues/16
 	delete envVars['RUBYGEMS_GEMDEPS'];
-	
+
 	var file,
+		args,
 		opts = {
 			cwd: verifierPath.cwd,
 			detached: !isWindows,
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -71,7 +71,6 @@
   "scripts": {
     "clean": "rimraf logs/*",
     "lint": "jscs .",
-    "pretest": "npm run clean && npm run lint",
```
mboudreau: Why'd you remove the linting?

mefellows: what you talking 'bout, it's right there... :)

mboudreau: You removed the pretest script, which is there to run the linting before the tests run.

mefellows: (it was a joke - I re-instated it. My original PR was on top of an outdated branch and forget to bring it back in).

mboudreau:
Seems to work, but will probably change code slightly before releasing.

mboudreau:
Also, might do the same thing for the pact mock service binary just in case.

mefellows:
Thanks @mboudreau, if you have feedback happy to take on board for future requests. I noticed an extra line break that could probably disappear. My editor automatically removes empty space, perhaps we should add a `.editorconfig`?

mboudreau:
Yeah, not a bad idea.

As per feedback, I'm just being pedantic and instead of having several
underscore functions nested, you can create a chain instead which makes it
more readable and easier to use, but that's just my preference :)

mefellows:
Sounds good to me!

PR #26, opened 2016-12-23
Added logging to fail handlers

marcodejongh:
I was having troubles getting my contracts published on CI, after much hair being pulled out I resorted to adding logs to publisher.js. Turned out it was failing because I made a typo in the provider name and the broker was returning the provider has not been created error.

To prevent other people from struggling with this I've added logs to the fail branches of the promises.

### DIFF
--- a/src/publisher.js
+++ b/src/publisher.js
@@ -22,6 +22,8 @@ function Publisher(pactBroker, pactUrls, consumerVersion, pactBrokerUsername, pa
 
 Publisher.prototype.publish = function () {
 	var options = this._options;
+	var errorMsg;
+	
 	logger.info('Publishing pacts to broker at: ' + options.pactBroker);
 
 	// Stat all paths in pactUrls to make sure they exist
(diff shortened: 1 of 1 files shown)


mboudreau:
Thanks for that, definitely a welcomed contribution.  My only request would be that if this is an error, could you please call `logger.error` instead of `logger.warn`.  After that, I'll merge it and release a new version.

Cheers.

mboudreau:
Hm, after some thought, I've started modifying how we're logging things.  In essence, I want to use the logger less and use promise resolutions instead to bubble up all errors.  I'm currently changing this right now, making your PR redundant.  I'll close it, but the fix should be coming up soon.

Cheers.

marcodejongh:
@mboudreau Sounds great, as long as it stops failing silently I'm happy 👍

mboudreau:
@marcodejongh agreed.  Silent fails suck :/

mboudreau:
@marcodejongh version 4.6.0 is out.  Please try it to make sure it solves your issue.  Cheers.

PR #35, opened 2017-05-09
feat(verifications): publish verifications to broker

mefellows:
Adds support for a new feature of the broker - the ability to send provider verification results back to the Broker, allowing more intelligent CI practices.

The main use case this solves is allowing you (probably via a CI / CD processe) to detect Consumer changes that have not yet been verified by a provider. Publishing the status to the broker allows tooling to then query this status, and block a deployment if the contract has changed or has not been verified since last published.

See https://github.com/realestate-com-au/pact/blob/master/CHANGELOG.md#1110-8-may-2017 for more.

**Commit details**:
- Add flags to automatically publish verification results to
  Pact broker
- Had to hard code port in integration test as verifier reads
  broker URL dynamically from pact (see publish-verification-example-{success,fail}.json)
- Bump minimum requirement for pact-provider-verifier to 1.1.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,8 @@ var opts = {
 	providerStatesSetupUrl: <String>, // URL to send PUT requests to setup a given provider state. Optional.
 	pactBrokerUsername: <String>,    // Username for Pact Broker basic authentication. Optional
 	pactBrokerPassword: <String>,    // Password for Pact Broker basic authentication. Optional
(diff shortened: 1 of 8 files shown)


mboudreau:
Looks good.

mefellows:
Thanks - shall I publish a new version?

mboudreau:
@mefellows I was in the process of doing it now as well as some quick upkeep that I've been meaning to do.

PR #36, opened 2017-05-10
Fetch from broker: Allow verifier to fetch pacts from a Broker

mefellows:
This feature allows a user to simply point at a Pact Broker via `pactBrokerUrl` with a Provider name (`provider`), optionally passing in a set of `tags`, and the library will automatically pull in all pacts for the given Provider.

This is the ideal way of finding pacts, rather than hard coding locations per consumer/version/tag etc.

The integration tests aren't fully featured, but they do show the underlying Ruby process is actually being properly executed which was my intention. The unit tests should do the job in terms of functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,15 +61,18 @@ Read more about [Verify Pacts](https://github.com/realestate-com-au/pact/wiki/Ve
 ```js
 var pact = require('@pact-foundation/pact-node');
 var opts = {
-	providerBaseUrl: <String>,       // Running API provider host endpoint. Required.
-	pactUrls: <Array>,               // Array of local Pact file paths or Pact Broker URLs (http based). Required.
-	providerStatesUrl: <String>,     // URL to fetch the provider states for the given provider API. Optional.
-	providerStatesSetupUrl: <String>, // URL to send PUT requests to setup a given provider state. Optional.
-	pactBrokerUsername: <String>,    // Username for Pact Broker basic authentication. Optional
-	pactBrokerPassword: <String>,    // Password for Pact Broker basic authentication. Optional
+	providerBaseUrl: <String>,           // Running API provider host endpoint. Required.
+	pactBrokerUrl: <String>              // URL of the Pact Broker to retrieve pacts from. Required if not using pactUrls.
+	provider: <String>                   // Name of the Provider. Required.
+	tags: <Array>                        // Array of tags, used to filter pacts from the Broker. Optional.
+	pactUrls: <Array>,                   // Array of local Pact file paths or HTTP-based URLs (e.g. from a broker). Required if not using a Broker.
+	providerStatesUrl: <String>,         // URL to fetch the provider states for the given provider API. Optional.
+	providerStatesSetupUrl: <String>,    // URL to send PUT requests to setup a given provider state. Optional.
+	pactBrokerUsername: <String>,        // Username for Pact Broker basic authentication. Optional
+	pactBrokerPassword: <String>,        // Password for Pact Broker basic authentication. Optional
 	publishVerificationResult: <Boolean> // Publish verification result to Broker. Optional
-	providerVersion: <Boolean>       // Provider version, required to publish verification result to Broker. Optional otherwise.
-	timeout: <Number>                // The duration in ms we should wait to confirm verification process was successful. Defaults to 30000, Optional.
+	providerVersion: <Boolean>           // Provider version, required to publish verification result to Broker. Optional otherwise.
+	timeout: <Number>                    // The duration in ms we should wait to confirm verification process was successful. Defaults to 30000, Optional.
 };
 
 pact.verifyPacts(opts).then(function () {
--- a/src/broker.js
+++ b/src/broker.js
@@ -1,0 +1,77 @@
+'use strict';
+
+var checkTypes = require('check-types'),
+	logger = require('./logger'),
+	traverson = require('traverson-promise'),
+	JsonHalAdapter = require('traverson-hal');
+
+// register the traverson-hal plug-in for media type 'application/hal+json'
+traverson.registerMediaType(JsonHalAdapter.mediaType, JsonHalAdapter);
+
+var pactURLPattern = "/pacts/provider/%s/latest",
+	pactURLPatternWithTag = "/pacts/provider/%s/latest/%s";
+
+// Constructor
+function Broker(brokerUrl, provider, tags, username, password) {
+	this._options = {};
+	this._options.brokerUrl = brokerUrl;
+	this._options.provider = provider;
+	this._options.tags = tags;
+	this._options.username = username;
+	this._options.password = password;
+}
+
+// Find all consumers
+Broker.prototype.findConsumers = function () {
+	logger.debug("Finding consumers")
+	return traverson
+		.from('https://test.pact.dius.com.au/pacts/provider/bobby/latest/sit4')
+		.withRequestOptions(this.getRequestOptions())
+		.jsonHal()
+		.follow('pacts')
+		.get()
+		.result;
+};
+
+Broker.prototype.getRequestOptions = function () {
+	if (this._options.username && this._options.password) {
+		return {
+			'auth': {
+				'user': this._options.username,
+				'password': this._options.password
+			}
+		}
+	}
+	return {}
+}
+
+
+
+// Creates a new instance of the Pact Broker HAL client with the specified option
+module.exports = function (options) {
+	options = options || {};
+
+	// defaults
+	options.brokerUrl = options.brokerUrl || '';
+	options.provider = options.provider || '';
+	options.username = options.username || '';
+	options.password = options.password || '';
+	options.tags = options.tags || [];
+
+	checkTypes.assert.nonEmptyString(options.brokerUrl);
+	checkTypes.assert.nonEmptyString(options.provider);
+
+	if (options.tags) {
+		checkTypes.assert.array.of.string(options.tags);
+	}
+
+	if (options.username) {
+		checkTypes.assert.string(options.username);
+	}
+
+	if (options.password) {
+		checkTypes.assert.string(options.password);
+	}
+
+	return new Broker(options.provider, options.brokerUrl, options.tags, options.username, options.password);
+};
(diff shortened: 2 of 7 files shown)


src/verifier.js
```diff
@@ -143,8 +163,28 @@ module.exports = function (options) {
 		throw new Error('Must provide both or none of --provider-states-url and --provider-states-setup-url.');
 	}
 
-	if (options.pactUrls) {
+	if (options.pactUrls && !checkTypes.emptyArray(options.pactUrls)) {
 		checkTypes.assert.array.of.string(options.pactUrls);
+		retrievePactsPromise = new Promise(function (resolve) {
```
mboudreau: Couldn't this be `Promise.resolve(options.pactUrls)` instead?

src/verifier.js
```diff
@@ -143,8 +163,28 @@ module.exports = function (options) {
...
+		var broker = require('./broker')({
+			brokerUrl: options.pactBrokerUrl,
+			provider: options.provider,
+			username: options.pactBrokerUsername,
+			password: options.pactBrokerPassword,
+			tags: options.tags
+		});
+		retrievePactsPromise = broker
+			.findConsumers()
+			.then(function (pactUrls) {
```
mboudreau: this `then` is useless.

src/verifier.js
```diff
@@ -22,81 +22,94 @@ function Verifier(providerBaseUrl, pactUrls, providerStatesUrl, providerStatesSe
...
 	options = options || {};
 	options.providerBaseUrl = options.providerBaseUrl || '';
+	options.pactBrokerUrl = options.pactBrokerUrl || '';
+	options.tags = options.tags || [];
+	options.provider = options.provider || '';
 	options.pactUrls = options.pactUrls || [];
 	options.providerStatesUrl = options.providerStatesUrl || '';
 	options.providerStatesSetupUrl = options.providerStatesSetupUrl || '';
 	options.timeout = options.timeout || 30000;
+	var retrievePactsPromise;
```
mboudreau: This promise, I'm not sure if it belongs here.  I think it would be better in the constructor instead of where we do all the options assertion.

src/broker.js
```diff
@@ -0,0 +1,100 @@
...
+	this._options.provider = provider;
+	this._options.tags = tags;
+	this._options.username = username;
+	this._options.password = password;
+};
+
+// Find Pacts returns the raw response from the HAL resource
+Broker.prototype.findPacts = function (tag) {
+	logger.debug("finding pacts: ", tag)
+	var linkName = (tag) ? 'pb:latest-provider-pacts-with-tag' : 'pb:latest-provider-pacts';
```
mboudreau: don't need the parentheses.

src/broker.js
```diff
@@ -0,0 +1,100 @@
...
+	logger.debug("Finding consumers")
+
+	var promises = (this._options.tags.length > 0) ? this._options.tags.map(this.findPacts, this) : [this.findPacts()];
+
+	return Promise
+		.all(promises)
+		.then(function (values) {
+			var pactUrls = {};
+			values.forEach(function (response) {
+				response._links.pacts.forEach(function (pact) {
```
mboudreau: do we know for sure that `response._links.pacts` won't have any nulls in it?

src/broker.js
```diff
@@ -0,0 +1,100 @@
...
+					pactUrls[pact.title] = pact.href;
+				})
+			});
+			return Object.keys(pactUrls).reduce(function (pacts, key) {
+				pacts.push(pactUrls[key])
```
mboudreau: should probably have some semi-colons to end the statement.

src/broker.js
```diff
@@ -0,0 +1,100 @@
...
+				return pacts
+			}, [])
+		})
+};
+
+Broker.prototype.getRequestOptions = function () {
+	if (this._options.username && this._options.password) {
```
mboudreau: does this object have to be 'built' at runtime?  Seems like it would be static and would never change.  I think it should be built within the constructor.

README.md
```diff
@@ -61,15 +61,18 @@ Read more about [Verify Pacts](https://github.com/realestate-com-au/pact/wiki/Ve
...
-	providerBaseUrl: <String>,       // Running API provider host endpoint. Required.
-	pactUrls: <Array>,               // Array of local Pact file paths or Pact Broker URLs (http based). Required.
-	providerStatesUrl: <String>,     // URL to fetch the provider states for the given provider API. Optional.
-	providerStatesSetupUrl: <String>, // URL to send PUT requests to setup a given provider state. Optional.
-	pactBrokerUsername: <String>,    // Username for Pact Broker basic authentication. Optional
-	pactBrokerPassword: <String>,    // Password for Pact Broker basic authentication. Optional
+	providerBaseUrl: <String>,           // Running API provider host endpoint. Required.
+	pactBrokerUrl: <String>              // URL of the Pact Broker to retrieve pacts from. Required if not using pactUrls.
+	provider: <String>                   // Name of the Provider. Required.
+	tags: <Array>                        // Array of tags, used to filter pacts from the Broker. Optional.
```
mboudreau: How's the array going to work for the CLI?

src/verifier.js
mefellows: Yikes, good spot.

src/broker.js
mefellows: Good call, this was left over from a refactor and is now redundant.

src/broker.js
mefellows: No, I'll fix this up and add a negative test case for empty response.

src/verifier.js
mefellows: Yes.

src/broker.js
mefellows: Thanks, done (surprisingly the lint didn't fail here - maybe we ought to create a tech-debt card for future stuff like this?)

README.md
mefellows: Hmmmm maybe comma separated and then split into an array?

But, does the CLI even work? Looking at `bin/pact-node` I can't see the `create()` method being referenced: `pact.create(program).start();`.

Running it results in:

```
matt ~/development/public/pact-node λ ./bin/pact-node
/Users/mfellows/development/public/pact-node/bin/pact-node:31
pact.create(program).start();
     ^

TypeError: pact.create is not a function
    at Object.<anonymous> (/Users/mfellows/development/public/pact-node/bin/pact-node:31:6)
    at Module._compile (module.js:541:32)
    at Object.Module._extensions..js (module.js:550:10)
    at Module.load (module.js:456:32)
    at tryModuleLoad (module.js:415:12)
    at Function.Module._load (module.js:407:3)
    at Function.Module.runMain (module.js:575:10)
    at startup (node.js:160:18)
    at node.js:445:3
```

I have been a bit blind to its existence so thanks for reminding me.

src/broker.js
```diff
@@ -19,16 +19,18 @@ function Broker(provider, brokerUrl, tags, username, password) {
...
+	this._requestOptions = (this._options.username && this._options.password) ? { 'auth': { 'user': this._options.username, 'password': this._options.password } } : {};
 };
 
 // Find Pacts returns the raw response from the HAL resource
 Broker.prototype.findPacts = function (tag) {
-	logger.debug("finding pacts: ", tag)
-	var linkName = (tag) ? 'pb:latest-provider-pacts-with-tag' : 'pb:latest-provider-pacts';
+	logger.debug("finding pacts for Provider:", this._options.provider, ", Tag:", tag)
+
+	var linkName = tag ? 'pb:latest-provider-pacts-with-tag' : 'pb:latest-provider-pacts';
```
mboudreau: Could be `var linkName = 'pb:latest-provider-pacts' + tag ? '-with-tags' : '';`. Might not even need to save it to a variable and just use it inline.

README.md
mboudreau: Huh, weird.  It was working before.  Must of been a refactor that didn't go through or something.  I need to get typescript going for it to fix some of these easily identifiable issues.  But yeah, if we can get it working again, that'd be great.  I think comma delimited would be good enough.

src/verifier.js
```diff
@@ -143,8 +163,22 @@ module.exports = function (options) {
 		throw new Error('Must provide both or none of --provider-states-url and --provider-states-setup-url.');
 	}
 
-	if (options.pactUrls) {
+	if (options.pactUrls && !checkTypes.emptyArray(options.pactUrls)) {
 		checkTypes.assert.array.of.string(options.pactUrls);
+		retrievePactsPromise = Promise.resolve(options.pactUrls);
```
mboudreau: Still think that this promise should be created within the constructor since this function is more about validating input.

mefellows: Yes, fair call. Actually moving it directly into the `verify()` function is even better (passing around promises here is a bit ugly). Will have an updated PR with this shortly.

mefellows:
Can we separate out the CLI uplift work from this PR? It's pretty big as it is, I'm happy to fix the CLI in the coming days?

P.S. I'd be in favour of the TS annotation work.

mboudreau:
Eh, you don't have to fix the CLI, but yeah, let's get it in a separate PR.

src/verifier.js
mefellows: Updated this btw, let me know what you think.

mboudreau:
@mefellows 

Seems we have a problem with a dependency with pact-mock-service.  Seems that version 2.0.1 doesn't exist.  Didn't someone unpublish it?

mboudreau:
@mefellows don't worry, deploying new version of pact-mock-service, will update pact-node with the latest

mefellows:
Thanks mate, I unpublished, but for some reason I couldn't republish (looks like it was cached or something). NPM registry showed old version and I didn't want to bump the version (as we got it all back in sync). I was going to do it this morning, but looks like you've sorted it - thanks so much!

mboudreau:
Yeah, NPM is weird around unpublishing because of an issue where someone
unpublished and killed a lot of projects that were dependent on it.  It's
much easier to just increment version number.

mefellows:
I remember that well - but I know how you have OCD like me and like to keep things in sync, I wasn't aware you couldn't re-publish.

Again, thanks man.

PR #40, opened 2017-06-12
Deprecate providerStatesUrl

mefellows:
Since the verifier `v1.1.3` the states URL is no longer required and is ignored - huzzah!

This PR removes the need for it and adds a deprecation warning.

Also, there was some minor formatting (spaces where there needed be any) in the verifier that my editor has kindly removed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "@pact-foundation/pact-node",
-	"version": "4.9.2",
+	"version": "4.10.0",
 	"description": "A wrapper for the Ruby version of Pact to work within Node",
 	"main": "./src/pact.js",
 	"os": [
@@ -37,8 +37,8 @@
 		"url": "https://github.com/pact-foundation/pact-node/issues"
 	},
 	"dependencies": {
-		"@pact-foundation/pact-mock-service": "~2.1.0",
-		"@pact-foundation/pact-provider-verifier": "~1.1.1",
+		"@pact-foundation/pact-mock-service": "^2.1.0",
+		"@pact-foundation/pact-provider-verifier": "^1.2.0",
 		"bunyan": "^1.8.10",
 		"bunyan-prettystream": "^0.1.3",
 		"check-types": "~7.1.5",
--- a/package.json
+++ b/package.json
@@ -37,8 +37,8 @@
 		"url": "https://github.com/pact-foundation/pact-node/issues"
 	},
 	"dependencies": {
-		"@pact-foundation/pact-mock-service": "^2.1.0",
-		"@pact-foundation/pact-provider-verifier": "^1.2.0",
+		"@pact-foundation/pact-mock-service": "~2.1.0",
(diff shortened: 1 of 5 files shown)


test/verifier.integration.spec.js
```diff
@@ -156,13 +156,7 @@ describe("Verifier Integration Spec", function () {
...
 							pactBrokerPassword: 'baaoeur'
 						});
-						return verifier.verify()
-							.then(function () {
-								throw new Error("Expected an error but got none");
-							})
-							.catch(function (err) {
-								expect(err.message).to.contain("Unauthorized")
-							})
+						return expect(verifier.verify()).to.eventually.be.rejected
```
mefellows: This was required as the behaviour slightly changed in the new verifier although functionally remained the same. This was much simpler and less fragile.

package.json
```diff
@@ -37,8 +37,8 @@
 		"url": "https://github.com/pact-foundation/pact-node/issues"
 	},
 	"dependencies": {
-		"@pact-foundation/pact-mock-service": "~2.1.0",
-		"@pact-foundation/pact-provider-verifier": "~1.1.1",
+		"@pact-foundation/pact-mock-service": "^2.1.0",
```
mboudreau: Please keep it for patch fixes only with the `~` instead of the `^`

mefellows: Up to you - it does cause a bit of a maintenance nightmare but the upside is control. Sure - want me to update?

mefellows: Also, build is green except for one mac osx build that times out. Need to get to the source of flakiness - could just be slowness starting the Ruby process.

package.json
```diff
@@ -38,7 +38,7 @@
 	},
 	"dependencies": {
 		"@pact-foundation/pact-mock-service": "~2.1.0",
-		"@pact-foundation/pact-provider-verifier": "~1.1.1",
+		"@pact-foundation/pact-provider-verifier": "~1.3.0",
```
mefellows: Have locked these back down @mboudreau, let me know if you want any more.

mboudreau:
Build is green, except for Node 5 for some reason.  Restarting to make sure it's just a one-off.

mboudreau:
Hm, seems that the verifier is a bit slow.  Might need to up the timeout on the tests since it seems that 2 seconds is too slow for the verifier?

mefellows:
It's green now!

I think the `--timeout 10000` is not working, given the error message is saying it's timing out after `2000ms`.

mefellows:
It's the way `cross-env` is handling it, it's not passing the flags through. Will investigate, but we're good for now I think.

mboudreau:
Cool :)

mefellows:
Looked at this over lunch. Popped a `mocha.opts` file in the `test` directory which is where all of the integration tests are. It will default to `2000ms` for the basic unit tests and increase to `10000ms` for the ones that need Ruby. This should make things a bit less flakey.

mboudreau:
Cool.  Ready to merge then :)

Issue #42, opened 2017-08-14
Not finding linux pact mock service on centos 7


soundstep:
I was previously using nodejs 4.3.2 on a centos 7 linux box running jenkins.
Since I've updated to nodejs 8.3.0, I can't seem to be able to require `@pact-foundation/pact-node`.
I get the following error:

```
Warning: Cannot find module '@pact-foundation/pact-mock-service-linux-x64' Use --force to continue.
```

I tried with `@pact-foundation/pact-node` version v4.8.1 and v4.12.0 (latest).

I first noticed the problem with karma-pact, but the problem happens in another repo where I don't use karma-pact.

soundstep:
Update, not related to nodejs v8.3.0 (unless it doesn't happen all the time).
I'm now able to load pact-node when I revert it to v4.8.1, so maybe just a problem with this v4.12.0.
Hope that helps.

soundstep:
Sorry another update, really hard to track what's going on.
I had the error once with nodejs v8.3.0 and and pact-node v4.8.1, so it is related to the latest nodejs version as I've never had it before, just not happening all the time.

The failure happens in a grunt task that looks like in case it is relevant:

```
grunt.registerTask('pact-service', 'Start pact mock service', function () {
        var done = this.async();
        var path = require('path');
        var pactWrapper = require('@pact-foundation/pact-node');
```

mefellows:
As you may have guessed, we use optional dependencies to pull down dynamically (install time) the correct OS-specific package, in this case linux 64 bit. It's common to see that warning, but of course if it doesn't install one of the packages things won't work!

Are you able to please enable detailed logging for the npm install step (forgotten how off the top of my head) and share a gist of the output? That might help us track it down.

FWIW we currently test Pact Nnode with whatever TravisCI provides, currently `v8.2.1`.

soundstep:
OS: Linux Centos 7
Nodejs: v8.3.0
NPM: 5.3.0
Pact-node: 4.8.1

You can find 2 log files in this gist, a npm install verbose and the jenkins job where the following error appears: 

```
Error: Cannot find module '@pact-foundation/pact-mock-service-linux-x64'
```

https://gist.github.com/soundstep/e8c1b72707fa18a041e591b7ed7b543b

Hope that helps!

mboudreau:
This is probably because if the way pact node tries to discover the os
specific version. I should have time to fix this tomorrow and try out node
8.

soundstep:
[...]

PR #44, opened 2017-08-23
Typescripting all the things!

mboudreau:
Still not completely done just yet, just a first pass.  Everything should work the same except I'm still trying to figure out the scenario where commonjs tries to require the library because of the convention of 'default' with typescript.


src/pact.spec.ts
```diff
@@ -0,0 +1,263 @@
...
+				pact.createServer({port: 1235});
+				pact.createServer({port: 1236});
+				return pact.removeAllServers()
+					.then(() => expect(pact.listServers()).to.be.empty);
+			});
+		});
+	});
+
+	// These tests never worked because the expect was wrong.  When fixed, massive issue ensues
+	xdescribe("Verify Pacts", () => {
```
mboudreau: @mefellows these 2 tests had to be excluded as they simply wasn't working.  The expect wasn't doing it's job because it was doing `to.be.eventually.resolved;` which is not an expect type, hence it was ignoring the error.  I'm not sure what you were trying to test specifically since the pactUrls isn't valid and causing issues with the verifier.

mboudreau:
@mefellows I still need to fix that one test that was failing, could you please try to remember what you were trying to test?  Also, can you do a quick review to see if it's all up to par?

mefellows:
Hoping to look at this tonight mate! I've been brushing up on typescript as a prerequisite so that any feedback is informed (at least, better informed).

mboudreau:
@mefellows Have you found some time yet?

mefellows:
I've started the review, it will take some time but I'll start feeding them in now. A lot will probably be just questions, but keen to get this in (as it will be great to have the types for the TS conversion of pact-js)

README.md
```diff
@@ -16,28 +16,36 @@ An idiomatic Node interface for the [Pact](http://pact.io) mock service (Consume
...
 var server = pact.createServer({port: 9999});
 server.start().then(function() {
 	// Do your testing/development here
 });
 ```
 
+Or you can use the CLI
+
+```
+$# pact mock --port 9999
```
mefellows: Does this actually work? If so, nice work!

bin/pact-cli.ts
```diff
@@ -0,0 +1,44 @@
...
+	.action((args, options) => pact.createServer(options).start());
+
+cli
+	.command("verify")
+	.description("Verifies Pact Contracts on the current provider")
+	.option("-b, --provider-base-url <URL>", "The Pact Provider base URL.")
+	.option("-u, --pact-urls [URLs]", "Comma separated list of Pact files or URIs.", cli.LIST)
+	.option("-p, --provider [name]", "Name of the Provider to verify. Required if not using --pact-urls.")
+	.option("-bu, --pact-broker-url [URL]", "URL of the Pact Broker to retrieve pacts from. Required if not using --pact-urls.")
+	.option("-pu, --provider-states-url [URL]", "URL to fetch provider states from.")
```
mefellows: I reckon we should drop this flag, it was deprecated and given this will likely be a major release it feels like a good time to make it disappear

bin/pact-cli.ts
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env node
```
mefellows: 👍

tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+	"compilerOptions": {
+		"module": "commonjs",
+		"target": "es5",
+		"noImplicitAny": false,
```
mefellows: Suggest we make this true.

src/broker.ts
```diff
@@ -0,0 +1,100 @@
...
+export class Broker {
+	public static create(options: BrokerOptions = {}) {
+		// defaults
+		options.brokerUrl = options.brokerUrl || "";
+		options.provider = options.provider || "";
+		options.username = options.username || "";
+		options.password = options.password || "";
+		options.tags = options.tags || [];
+
+		checkTypes.assert.nonEmptyString(options.brokerUrl);
```
mefellows: Could we not use stronger typing here in the `BrokerOptions` to avoid checking all the types etc.?

src/broker.ts
```diff
@@ -0,0 +1,100 @@
...
+				}, []);
+			})
+			.catch(() => q.reject(`Unable to find pacts for given provider '${this.__options.provider}' and tags '${this.__options.tags}'`));
+	}
+}
+
+// Creates a new instance of the Pact Broker HAL client with the specified option
+export default Broker.create;
+
+export interface BrokerOptions {
```
mefellows: As per above, could perhaps the mandatory fields here be made mandatory?

src/logger.ts
```diff
@@ -0,0 +1,32 @@
...
+
+	public get logLevelName():string {
+		return bunyan.nameFromLevel[this.level()];
+	}
+}
+
+export type LogLevels = "trace" | "debug" | "info" | "warn" | "error" | "fatal";
+
+export default new Logger({
+	name: `search-api@${pkg.version}`,
```
mefellows: What is `search-api`?

mefellows:
I've submitted some comments just now, next is to test locally with `pact-js` for compatibility etc., but I'll get stuck into some more over the next few days. Really great work.

bin/pact-cli.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+					return q.reject(new Error("Pact stop failed; tried calling service 10 times with no result."));
+				}
+				// Try again in 1 second
+				return q.delay(1000).then(() => this.checkMockStarted(options, amount));
+			});
+	}
+
+	private static call(options: ServerOptions): q.Promise<any> {
+		// console.log("Calling to see if pact service is up");
+		options.ssl = options.ssl || false;
```
mefellows: Do we need to do this? Shouldn't TS automatically set them to false/defaults given their explicit type?

mboudreau: Yes, we do, because you still need to remember that TS just compiles to javascript, which doesn't have type.  It works fine if someone is using TS with pact-node, but won't work well if someone is using javascript with it.

mefellows: True, it feels like we still have to do a lot of dancing because TypeScript _doesn't_ do this when transpiled back to JS.

mboudreau: there is a library that does this for you, but I've yet to try it.

mefellows:
Just ran the regression suites on pact-js linked against the new pact-node and all ✅

bin/pact-cli.spec.ts
mefellows: Which libraries out of interest?

.gitignore
```diff
@@ -34,3 +34,18 @@ node_modules
 *.iml
 
 .tmp
+
+# npm
+package-lock.json
```
mefellows: I meant to ask about this, I'm guessing this has been excluded due to #42?

bin/pact-cli.spec.ts
mboudreau: few options here: https://github.com/Microsoft/TypeScript/issues/1573

As well as https://www.npmjs.com/package/runtime-type-checks and https://github.com/gcanti/io-ts

Haven't played with any of them, but would remove the need for any type checking in our own code and let something else handle it all.

.gitignore
mboudreau: Yes and no.  On purpose, I exclude in all my projects, the reason being is that NPM did not think it through properly and actually broke optional dependencies, but also, I simply do not see the actual purpose if it.  The package.json file has the ability to "lock" the dependency to a particular version number, which is something is constantly do for production systems so that everyone should always have the same version of the libraries, especially the CI system.  Package.lock is just a way to get around semver, which is a blessing and a curse, but doesn't really add any benefit to a project unless people like to play fast and loose with the versioning.

bin/pact-cli.ts
mboudreau: Which flag specifically? the provider states one?

tsconfig.json
mboudreau: Can't as we do have any in our code, plus, it means all dependencies will also need to follow this rule, which is problematic.  Nothing wrong with using any, just don't want to use it abundantly.

src/broker.ts
mboudreau: Will need to add runtime type checking eventually.

src/broker.ts
mboudreau: Good catch, changing that now.

src/logger.ts
mboudreau: whoops, copy pasted that from another project :P

mboudreau:
@mefellows as for your comments:

- Unable to run npm run build: do a `git clean -xdf` then do `npm install` again and it should work.  I'm guessing a dependency didn't download properly
- VS Code is actually quite horrible compared to a good browser like WebStorm :P Either way, VS Code has difficulties following any kind of coding conventions since it seems to ignore tsling.conf and doesn't have .editorconfig support out of the box, need to install an extension to do that.  Essentially, it comes down to, follow TSlint and everything should be good :)
- require vs import: import has limitations, as in, it has to be at the start of the file, so dynamically importing things is out of the question (this can be good or bad depending on how you look at it), personally, I tend to use require for javascript imports and import for typescript.  There are some javascript things that don't seem to work well with how import works under the hood.
- Nothing is "normal" in the javascript world :P There's no defined convention that I can see as standard.  I have in the past added compiled code to a different directory, but it then makes it harder to package/deploy things since you then have to move key files (like package.json) to the directory as well.  In this particular case, it seemed easier to just leave it side by side for deployment, but also helps with debugging the compiled code if needed.  If you were to use a good IDE, it would collapse those generated files under the TS file so you don't see them :)
- Microsoft coding guidelines doesn't seem to mention as to why to not use them.  I used them because [TS doesn't minify the compiled js or mangle private properties or functions, making them still completely available in compiled javascript under the same name]given(https://github.com/Microsoft/TypeScript/issues/8).  By using an underscore (I prefer double underscore) convention for privates, we can set uglify to minify and mangle any property starting with `__`.  I've used this with great success in another web project and it reduced the file size considerably.

bin/pact-cli.ts
mboudreau: ah, the provider-states-url one.  Removing.

mboudreau:
@mefellows anything else?

mefellows:
@mboudreau nope, I think that's good. I saw a commit yesterday relating to a JS interop test - I think that's a good idea and will reduce need for other integration style tests in other repos.

I'll be doing the same in `pact-js`. 👍

src/server.ts
```diff
@@ -0,0 +1,348 @@
...
+		if (this.__instance) {
+			pid = this.__instance.pid;
+			logger.info(`Removing Pact with PID: ${pid}`);
+			this.__instance.removeAllListeners();
+			// Killing instance, since windows can't send signals, must kill process forcefully
+			try {
+				if (isWindows) {
+					cp.execSync(`taskkill /f /t /pid ${pid}`);
+				} else {
+					process.kill(-pid, "SIGINT");
```
mefellows: What's the reason for the `-pid` rather than just `pid`?

mboudreau: it's how unix system does sigint, but specifying the negative PID.  Not sure why, but it works so...

mefellows:
So excited to see this in, next up pact-js!

🎉   🚀  :shipit:

mboudreau:
@mefellows bit of a backlog of builds on travis, so might take a bit of time for it to go out, but yeah, the excitement is palpable :)

PR #45, opened 2017-09-22
feat(standalone): implement standalone pact service

mefellows:
Removes the two npm packages in favour of a single standalone package.

1. Means faster time-to-market (the rest of the languages are using this single package, so it will save us from having to publish all of the gems, then the two packages etc. etc.)
2. Means smaller distribution, as there is only a single Ruby runtime for both the mock service and verifier.

See https://travis-ci.org/pact-foundation/pact-node/jobs/278567944 for a passing build (it's currently red, but that's because I pushed this branch before I published the standalone module).

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,8 +37,7 @@
 		"url": "https://github.com/pact-foundation/pact-node/issues"
 	},
 	"dependencies": {
-		"@pact-foundation/pact-mock-service": "~2.1.0",
-		"@pact-foundation/pact-provider-verifier": "^1.4.3",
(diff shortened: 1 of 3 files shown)


mefellows:
Also, I'm totally aware of the Typescript work going on. If you're in theory happy with this, I'll happily port to the TS bit (it's pretty trivial after all).

mboudreau:
Using TS implementation instead, gird your loins, we're going in.

mefellows:
Consider my loins... girded.

PR #46, opened 2017-09-27
feat(standalone): add pact standalone to TS implementation

mefellows:
See https://github.com/pact-foundation/pact-node/pull/45 for the non-TS implementation of the same feature.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -167,7 +167,7 @@ export class Server extends events.EventEmitter {
 			"provider": "--provider"
 		});
 
-		let cmd: string = [pact.mockServicePath].concat(args).join(" ");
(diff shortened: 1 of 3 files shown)


src/server.ts
```diff
@@ -167,7 +167,7 @@ export class Server extends events.EventEmitter {
 			"provider": "--provider"
 		});
 
-		let cmd: string = [pactPath.file].concat(args).join(" ");
+		let cmd: string = [pact.mockServicePath].concat("service", ...args).join(" ");
```
mefellows: See https://github.com/pact-foundation/pact-python/issues/40 and https://github.com/pact-foundation/pact-python/commit/4661406b235b3cb5fe5a6013514d19da9f43bd1d.

We already track the PID so this change was a bit easier.

PR #53, opened 2017-11-05
feat(types): add TS declarations to package.json. Fixes #52

### LINKED ISSUE #52: Typescript declarations not found
When trying to upgrade `pact-node` in the `pact` module and removing the custom typings I had results in the following issue (truncated for readability):

```
src/pact-web.ts(20,28): error TS7016: Could not find a declaration file for module '@pact-foundation/pact-node'. '/Users/mfellows/development/public/pact-js/node_modules/@pact-foundation/pact-node/src/index.js' implicitly has an 'any' type.
  Try `npm install @types/@pact-foundation/pact-node` if it exists or add a new declaration (.d.ts) file containing `declare module '@pact-foundation/pact-node';`
```

I'm running the latest TypeScript (v2.6.1 at this time). Even though the advice from https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html says:

> Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so.

Manually adding `"types": "./src/index.d.ts",` to `package.json` solves the issue.

Without this, I must use the non-TS `const pact = require('...')` declaration, losing the nice auto-completion etc. and benefits of TS.

mefellows:


mboudreau:
@mefellows I'll be deploying this in a minute :)

mefellows:
Thanks!

PR #60, opened 2017-11-10
Added --monkeypatch option to pact mock service start up command and …

### LINKED ISSUE #54: Please add --monkeypatch option to pact mock service start up command and pact-provider-verifier command
@bethesque commented on [Tue Nov 07 2017](https://github.com/pact-foundation/pact-js/issues/123)

Requires standalone version pact-1.19.0.

This is to support: https://github.com/pact-foundation/pact-mock_service/issues/78 and another issue that I can't currently find.

The feature that the user is requesting is not something that should be changed for all users, as it is a non-standard behaviour, and breaks the Rack spec. However, we can allow the underlying code to be monkey patched for his specific use case. The value supplied to `--monkeypatch` will be an absolute path to a ruby file that will monkeypatch the underlying header logic.

This is not something that we need to document, but we can use it to deal with edge cases when necessary!

---

@mefellows commented on [Tue Nov 07 2017](https://github.com/pact-foundation/pact-js/issues/123#issuecomment-342373035)

Moving this to `pact-node` as that is where the the CLI is integrated.

oliver-gramberg:
…pact provider verifier command

Fix #54 

These changes have been made blind, as I have not yet been able to set up an environment where I can compile and test (help appreciated), but they have been reviewed by our local TS expert :-)


oliver-gramberg:
"when user specifies valid options
        1) should return serverFactory using specified options"

fails with

  "1) Pact Spec Create serverFactory when user specifies valid options should return serverFactory using specified options:
     Error: Monkeypatch not found at path: /absolute/path/to/ruby/patch.rb
      at Object.Server.create [as default] (src\server.ts:130:11)
      at Pact.createServer (src\pact.ts:30:29)
      at Context.<anonymous> (src\pact.spec.ts:94:23)"

because the named file obviously doesn't exist. Not sure how to specify a file such that it will be found in the builds.

oliver-gramberg:
Now, 

"when user specifies invalid log
        1) should return an error on invalid path"

fails with

  1) Pact Spec Create serverFactory when user specifies invalid log should return an error on invalid path:
     AssertionError: expected [Function] to throw Error
      at Context.<anonymous> (src\pact.spec.ts:168:16)"

The origin version of the test,

```
pact.createServer({log: path.resolve(dirPath, "log.txt")});
expect(fs.statSync(dirPath).isDirectory()).to.be.true;
```

seemed not to do what its spec said, so I tried to fix it with,

```
expect(() => {
	pact.createServer({log: path.resolve(dirPath, "log.txt")});
}).to.throw(Error);
```

but since it does

```
try {
	fs.statSync(path.normalize(options.dir)).isDirectory();
} catch (e) {
	mkdirp.sync(path.normalize(options.dir));
}
```

it never fails - unless I find a name that is invalid in all possible environments. Linux only forbids '/' in file/dir names, but any '/' I use will be swallowed by path.normalize().

I propose to delete this test. Will comment it out for now.

bethesque:
Yes, I think the mock just creates whatever path you give it for the log.

mefellows:
Thanks for this @oliver-gramberg, it mostly looks OK to me. There's another fundamental PR in the works at the moment, this is probably best done after that.

bethesque:
Can you get the PWD via a function and prepend it to the relative path of the monkeypatch file to get its absolute path? I would actually expect there to be an absolute path function in the file utils class.

oliver-gramberg:
@bethesque, sorry not sure where exactly you want me to do that. I did not find the place where the CLI is being called. Different repo?

bethesque:
> Not sure how to specify a file such that it will be found in the builds.

I was suggesting a solution for that. Something like `"${Dir.pwd}/relative/path/to/file.rb"`.

Can I make a suggestion here? You seem to be testing the functionality, not the "contract" (interface). All your test needs to do is show that the absolute path is passed correctly to "--monkeypatch", rather than prove that the behaviour is correct. Along the lines of: https://docs.pact.io/best_practices/contract_tests_not_functional_tests.html

bethesque:
> I did not find the place where the CLI is being called.

I'm not actually sure how it works, as I don't maintain the js repos, sorry.

oliver-gramberg:
OK, it's about the test code, not the production code. That answers my "where" question, at least partly. I was assuming the opposite, thinking you wanted me to actually get the monkeypatch parameter set.

Always happy to learn! I like the argument the article you referenced makes, and I think I will actually be able to apply it my current project. For this fork/PR: I copied existing test cases and adapted them to use the monkeypatch parameter. Yes, it's testing the functionality (of the wrapper code), but so are the tests where I copied from, and I am not testing deeper than those. Or am I? If I were to test the correct implementation of the monkeypatch, then shouldn't I have coded an expectation about the non-mangling of HTTP headers?

I am still not sure we're talking about the same pieces of code. Could we use the commenting feature of Github on the code diff tab of the PR?

src/verifier.spec.ts
```diff
@@ -73,6 +73,18 @@ describe("Verifier Spec", () => {
...
+				expect(() => {
+					verifierFactory({
+						providerBaseUrl: "http://localhost",
+						pactUrls: ["http://idontexist"],
+						monkeypatch: "/this/file/does/not/exist.rb"
+					});
+				}).to.throw(Error);
+			});
+		});
+
```
bethesque: The CLI can really only return a couple of responses. It can return an exit code 0, or an exit code 1. The only thing you can do if you get an error response is to spit out the error message, and fail the calling code. Personally, I wouldn't have this test, because all the "how I handle the exit error code" tests are going to be the same. All you need to test is "how should the client behave when it gets an error", not "what are all the different things that can cause an error". Theoretically, the underlying Ruby process could change its implementation to just ignore a missing monkeypatch file, and then this test would fail, so it's not actually giving you any feedback on the javascript code.

bethesque:
I've added a comment next to the bit of code that made me think of the "test the interface, not the behaviour" approach.

src/verifier.spec.ts
oliver-gramberg: This test exercises src/server.ts:125 ff. where I check if the named file exists. If not, the CLI is never called, but an Error is thrown.

mefellows:
Bump, what are we thinking on this issue folks?

bethesque:
I would like it in please. We don't need to advertise the feature, but it helps us work around limitations of certain libraries (eg. Rack::Test which expects headers to always be in Dasherized-Capitalized-Format)

mboudreau:
I'm really not a fan of fixing this level since it is something inherently wrong with the core library, hence any language specific implementation that uses it will break.  Would be more efficient to fix it at the core.

bethesque:
`it is something inherently wrong with the core library` - there is an assumption in Rack::Test that all HTTP headers follow the same format. While it is recommended and common to use capitalized words with dashes, it is not actually mandatory, and there is at least one person out there who can't use Pact because of this code, as they use lower case and underscores. I am not in a position to change the behaviour of Rack::Test, as it is the most widely used HTTP testing library in Ruby, and for 99% of use cases, they are correct to make the assumption that they do. I can, however, provide a way for the probably less than 1% of users to alter the code to allow them to use our library.

mboudreau:
Sure, I understand that, but again, all libraries that are using the core
will have this issue, so instead of fixing it only in pact-node, how about
we fix it closer to the root cause?

bethesque:
Are you suggesting that I fix the Ruby code in the standalone? As I've said, I can't fix it. I can only provide a way to load a Ruby file that can alter the behaviour. To do this, I have added a `--monkeypatch file` argument to the Ruby standalone CLI. This PR is just exposes that option to the javascript code.

mefellows:
@bethesque just thinking out loud here, could this be resolved by overriding the headers with the the `--custom-provider-header` flag? 

Down the track, supporting "request filters" as a first class citizen could also be a nice catch here.

bethesque:
I spent quite a lot of time looking in to this, and this was the only solution I could come up with.

bethesque:
Here's an example of why this feature is really useful. I've been trying to fix [this](https://github.com/pact-foundation/pact-js/issues/150) issue. The code that needs to be altered is embedded deep in the Ruby chain. By using the [monkeypatch](https://github.com/pact-foundation/pact-ruby-standalone-windows-test/commit/0352de0a464f00114c062cb0310585d67b1eef71) file, I can modify the code to allow me to test potential fixes without having to repackage everything (ruby, travelling ruby, js...) every time. Fast feedback loops lead to fast bug fixes. By exposing the monkeypatch option in the js, I'll be able to experiment with fixes really easily, or allow other people to try out fixes on their local machines if it's something that I can't reproduce (which seems to happen a lot with windows!)

mboudreau:
Alright, fine, I'll get this done. Should we document this feature with an
extreme warning or not document it at all so people don't abuse it?

bethesque:
Let's not document it. It'll only need to be used under "adult supervision" (ie. instructions from me!)

mboudreau:
cool :)

mboudreau:
I'm going to close this PR as I merged it manually because of all the major changes that had happened since and added more tests to make sure that the monkeypatch option actually works (thanks @bethesque for the mini test with the ruby file).

https://github.com/pact-foundation/pact-node/commit/8701b4db9bc5fd50bdc1e02cd076e2fe490a77ea

Version 6.11.0 will be going out soon with this.

PR #62, opened 2017-11-20
Improve type checking across codebase

### LINKED ISSUE #55: Improve compiler validations to support use as library
As `pact-node` is a library (in addition to a tool) and will be embedded in potentially many others, we have a duty to ensure the highest levels of TS language use.

An annoying restriction of TS is that it will compile, and therefore complain, about upstream dependencies.

e.g. for `pact-js` even with exclusions set as follows:

```
  "include": [
    "./src/**/*"
  ],
  "exclude": [
    "./node_modules/**/*",
    "./dist",
    "./examples",
    "**/*.spec.ts"
  ]
```

TS still picts up `pact-node`, as it is in compilation scope, so all compiler options for the external project are applied equally to `pact-node`, resulting in lots of errors (I disabled the `noImplicitAny` to reduce noise):

```
node_modules/@pact-foundation/pact-node/src/broker.ts(70,21): error TS2532: Object is possibly 'undefined'.
node_modules/@pact-foundation/pact-node/src/broker.ts(70,53): error TS2532: Object is possibly 'undefined'.
node_modules/@pact-foundation/pact-node/src/broker.ts(81,17): error TS2345: Argument of type 'any' is not assignable to parameter of type 'never'.
node_modules/@pact-foundation/pact-node/src/logger.ts(11,8): error TS2339: Property 'info' does not exist on type 'Logger'.
node_modules/@pact-foundation/pact-node/src/logger.ts(19,36): error TS2339: Property 'level' does not exist on type 'Logger'.
node_modules/@pact-foundation/pact-node/src/logger.ts(25,16): error TS2554: Expected 0 arguments, but got 1.
node_modules/@pact-foundation/pact-node/src/pact-util.ts(7,9): error TS7030: Not all code paths return a value.
node_modules/@pact-foundation/pact-node/src/publisher.ts(100,20): error TS2345: Argument of type 'any' is not assignable to parameter of type 'never'.
node_modules/@pact-foundation/pact-node/src/server.ts(243,4): error TS2322: Type 'undefined' is not assignable to type 'ChildProcess'.
node_modules/@pact-foundation/pact-node/src/verifier.ts(31,25): error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.
node_modules/@pact-foundation/pact-node/src/verifier.ts(112,7): error TS2532: Object is possibly 'undefined'.
src/dsl/verifier.ts(8,25): error TS2339: Property 'verifyPacts' does not exist on type 'typeof "/Users/mfellows/development/public/pact-js/node_modules/@pact-foundation/pact-node/src/in...'.
src/pact.ts(59,34): error TS2339: Property 'createServer' does not exist on type 'typeof "/Users/mfellows/development/public/pact-js/node_modules/@pact-foundation/pact-node/src/in...'.
src/pact.ts(70,20): error TS2339: Property 'logLevel' does not exist on type 'typeof "/Users/mfellows/development/public/pact-js/node_modules/@pact-foundation/pact-node/src/in...'.
```

My suggestion is that we increase strictness/validation compilation rules, but I'm open to other alternatives.

mefellows:
Adds the following TS configuration, in order to improve the libraries portability into other TS projects:

* --noImplicitAny: Raise error on expressions and declarations with an implied any type.
* --noImplicitReturns: Report error when not all code paths in function return a value.
* --noImplicitThis: Raise error on this expressions with an implied any type.
* --strictNullChecks: In strict null checking mode, the nulland undefined values are not in the domain of every type and are only assignable to themselves and any (the one exception being that undefined is also assignable to void).

NOTE: in a few q-related cases, explicit type casting was necessary. I haven't been able to dig far enough into why the code works but types are incorrect - either there is an issue with the code, or the types aren't quite correct (in particular, a `q.Resolve(...)` is not compatible with the `PromiseLike` interface).

Fixes #55.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,3 +14,7 @@ trim_trailing_whitespace = false
 [*.js]
 indent_style = tab
 indent_size = 4
+
+[package.json]
+indent_style = space
+indent_size = 2
--- a/bin/pact-cli.spec.ts
+++ b/bin/pact-cli.spec.ts
@@ -3,11 +3,11 @@ import chaiAsPromised = require("chai-as-promised");
 import childProcess = require("child_process");
 import q = require("q");
 import path = require("path");
-import _ = require("underscore");
 import {ChildProcess} from "child_process";
 import {ServerOptions} from "../src/server";
-import decamelize = require("decamelize");
 import providerMock from "../test/integration/provider-mock";
+const decamelize = require("decamelize");
+const _ = require("underscore");
 
 const request = q.denodeify(require("request"));
 const pkg = require("../package.json");
@@ -64,7 +64,7 @@ describe("Pact CLI Spec", () => {
 		});
 
 		context("with mock broker", () => {
-			let server;
+			let server: any;
 			const PORT = 9123;
 			const providerBaseUrl = `http://localhost:${PORT}`;
 
@@ -83,7 +83,7 @@ class CLI {
 	public static runMock(options: ServerOptions = {}): q.Promise<CLI> {
 		const args = _.chain(options)
 			.pairs()
-			.map((arr) => [`--${decamelize(arr[0], "-")}`, `${arr[1]}`])
+			.map((arr: any[]) => [`--${decamelize(arr[0], "-")}`, `${arr[1]}`])
 			.flatten()
 			.value();
 
@@ -113,6 +113,7 @@ class CLI {
 					cp.process.once("exit", () => deferred.resolve());
 					return deferred.promise;
 				}
+				return null;
 			});
 	}
 
--- a/src/broker.ts
+++ b/src/broker.ts
@@ -1,5 +1,6 @@
 import q = require("q");
 import logger from "./logger";
+import { IWhenable } from "q";
 const _ = require("underscore");
 const checkTypes = require("check-types");
 const request = q.denodeify(require("request"));
@@ -50,7 +51,7 @@ export class Broker {
 			} : null
 		};
 		return request(requestOptions)
-			.then((data) => data[0])
+			.then((data: any) => data[0])
 			.then((response) => {
 				if (response.statusCode < 200 && response.statusCode >= 300) {
 					return q.reject(response);
@@ -58,7 +59,7 @@ export class Broker {
 				const body = JSON.parse(response.body);
 				return request(_.extend({}, requestOptions, {uri: body._links[`pb:latest-provider-pacts${tag ? "-with-tag" : ""}`].href.replace("{tag}", tag).replace("{provider}", this.options.provider)}));
 			})
-			.then((data) => data[0])
+			.then((data: any) => data[0])
 			.then((response) => response.statusCode < 200 && response.statusCode >= 300 ? q.reject(response) : JSON.parse(response.body));
 	}
 
@@ -75,7 +76,7 @@ export class Broker {
 				}
 				return array;
 			}, []))
-			.catch(() => q.reject(`Unable to find pacts for given provider '${this.options.provider}' and tags '${this.options.tags}'`));
+			.catch(() => q.reject(`Unable to find pacts for given provider '${this.options.provider}' and tags '${this.options.tags}'`) as IWhenable<any>);
 	}
 }
 
--- a/src/broker.ts
+++ b/src/broker.ts
@@ -76,7 +76,7 @@ export class Broker {
 				}
 				return array;
 			}, []))
-			.catch(() => q.reject(`Unable to find pacts for given provider '${this.options.provider}' and tags '${this.options.tags}'`) as IWhenable<any>);
+			.catch(() => q.reject<IWhenable<any>>(`Unable to find pacts for given provider '${this.options.provider}' and tags '${this.options.tags}'`));
 	}
 }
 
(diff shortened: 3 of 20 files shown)


.editorconfig
```diff
@@ -14,3 +14,7 @@ trim_trailing_whitespace = false
 [*.js]
 indent_style = tab
 indent_size = 4
+
+[package.json]
```
mefellows: This should prevent that reformatting ;)

src/broker.ts
```diff
@@ -51,32 +51,32 @@ export class Broker {
...
 		return q.all(promises)
-			.then((values) => _.reduce(values, (array, v) => {
+			.then((values) => _.reduce(values, (array: string[], v: any) => {
 				if (v && v._links && v._links.pacts) {
 					array.push(..._.pluck(v._links.pacts, "href"));
 				}
 				return array;
 			}, []))
-			.catch(() => q.reject(`Unable to find pacts for given provider '${this.options.provider}' and tags '${this.options.tags}'`));
+			.catch(() => q.reject(`Unable to find pacts for given provider '${this.options.provider}' and tags '${this.options.tags}'`) as IWhenable<any>);
```
mefellows: This is an example of that `q` problem - a `q.Promise` is not compatible with a `IWhenable<T>` which it thinks a `q.all(...)` returns. 

Another potential option here is to modify the `q` import to not import the types, but you do lose out on some of the nice code completion and type completeness elsewhere in the file.

mefellows:
Any thoughts @mboudreau or should we merge this in (no API compatibility changes).

mboudreau:
@mefellows Ah, right, I'll look at this today.

src/broker.ts
mboudreau: isn't that the reason why q support generics?

bin/pact-cli.spec.ts
```diff
@@ -3,11 +3,11 @@ import chaiAsPromised = require("chai-as-promised");
 import childProcess = require("child_process");
 import q = require("q");
 import path = require("path");
-import _ = require("underscore");
 import {ChildProcess} from "child_process";
 import {ServerOptions} from "../src/server";
-import decamelize = require("decamelize");
 import providerMock from "../test/integration/provider-mock";
+const decamelize = require("decamelize");
```
mboudreau: isn't import and const the same thing in this case?

src/pact.spec.ts
```diff
@@ -5,14 +5,14 @@ import fs = require("fs");
...
-
 chai.use(chaiAsPromised);
+const currentDir = (process && process.mainModule) ? process.mainModule.filename : "";
 
 describe("Pact Spec", () => {
 	afterEach(() => pact.removeAllServers());
 
 	describe("Set Log Level", () => {
-		let originalLogLevel;
+		let originalLogLevel: any;
```
mboudreau: string

src/pact.spec.ts
```diff
@@ -51,7 +51,7 @@ describe("Pact Spec", () => {
 	});
 
 	describe("Create serverFactory", () => {
-		let dirPath;
+		let dirPath: any;
```
mboudreau: string

src/pact.ts
```diff
@@ -32,7 +32,7 @@ export class Pact {
 		logger.info(`Creating Pact Server with options: \n${this.__stringifyOptions(server.options)}`);
 
 		// Listen to server delete events, to remove from server list
-		server.once("delete", (s) => {
+		server.once("delete", (s: any) => {
```
mboudreau: Pact

src/server.spec.ts
```diff
@@ -12,7 +12,7 @@ chai.use(chaiAsPromised);
 const expect = chai.expect;
 
 describe("Server Spec", () => {
-	let server;
+	let server: any;
```
mboudreau: Pact

src/server.spec.ts
```diff
@@ -57,7 +57,7 @@ describe("Server Spec", () => {
 		});
 
 		context("when valid options are set", () => {
-			let dirPath;
+			let dirPath: any;
```
mboudreau: string

mefellows: 👍

src/server.spec.ts
mefellows: I think you mean `Server`. Fixing.

src/pact.spec.ts
mefellows: 👍

src/pact.spec.ts
mefellows: Actually it's a union type of `LogLevels` and `number | void`, as it gets use to both set and retrieve the log level - hence why I put any.

bin/pact-cli.spec.ts
mefellows: No, `const` doesn't try to resolve the module whereas `import` does. e.g. in this case the following error appears when using `import`:

```sh
bin/pact-cli.spec.ts(9,29): error TS7016: Could not find a declaration file for module 'decamelize'. '/Users/mfellows/development/public/pact-node/node_modules/decamelize/index.js' implicitly has an 'any' type.
  Try `npm install @types/decamelize` if it exists or add a new declaration (.d.ts) file containing `declare module 'decamelize';`
```

PR #65, opened 2017-12-07
Add Stub Service feature

mefellows:
* Add a stub service feature
* Refactor mock service to use common `AbstractService` base class (stub service shares a lot of functionality)
* Add stub service documentation
* Refactor documentation with TOC and table-based sections for parameter descriptions


mefellows:
Any chance you've had a moment to look at this @mboudreau? Just merged latest from master and fixed a minor issue.

mboudreau:
yep, sorry, been full out for the last few weeks :/

mefellows:
I've surprisingly had pockets of time to be able to contribute - wanted to make hay while the sun shines :)

mboudreau:
@mefellows quick question, what does stub do?  Is it just a mock server but without having to validate the interaction so it can be used for external services that you can't control/test on?

README.md
```diff
@@ -10,6 +10,32 @@
 
 An idiomatic Node interface for the [Pact](http://pact.io) mock service (Consumer) and Verification (Provider) process.
 
+<!-- TOC -->
```
mboudreau: Nice touch :)

src/pact.ts
```diff
@@ -56,6 +58,43 @@ export class Pact {
...
+	}
+
+	// Return arrays of all stubs
+	public listStubs(): Stub[] {
+		return this.__stubs;
+	}
+
+	// Remove all the stubs that have been created
+	// Return promise of all others
+	public removeAllStubs(): q.Promise<Stub[]> {
```
mboudreau: For simplicity in testing, I think we should add a `removeAll` function that simply removes all servers and stubs.

src/pact.ts
```diff
@@ -1,12 +1,14 @@
 import q = require("q");
 import serverFactory, {Server, ServerOptions} from "./server";
+import stubFactory, {Stub, StubOptions} from "./stub";
 import verifierFactory, {VerifierOptions} from "./verifier";
 import publisherFactory, {PublisherOptions} from "./publisher";
 import logger, {LogLevels} from "./logger";
 import _ = require("underscore");
```
mboudreau: should be `const`

src/server.ts
```diff
@@ -1,13 +1,10 @@
 // tslint:disable:no-string-literal
+import { AbstractService } from "./service";
```
mboudreau: Neat :)

src/server.ts
```diff
@@ -16,68 +13,11 @@ const CHECKTIME = 500;
 const RETRY_AMOUNT = 60;
```
mboudreau: Constants not being used.

src/pact.ts
```diff
@@ -56,6 +58,43 @@ export class Pact {
...
+			logger.error(msg);
+			throw new Error(msg);
+		}
+
+		let stub = stubFactory(options);
+		this.__stubs.push(stub);
+		logger.info(`Creating Pact Stub with options: \n${this.__stringifyOptions(stub.options)}`);
+
+		// Listen to stub delete events, to remove from stub list
+		stub.once("delete", (s: Stub) => {
```
mboudreau: should be using `Stub.Events.DELETE`

src/server.ts
```diff
@@ -127,9 +51,7 @@ export class Server extends events.EventEmitter {
 	}
 
 	public readonly options: ServerOptions;
-	private __running: boolean;
-	private __instance: ChildProcess;
-	private readonly __argMapping = {
+	protected readonly __argMapping = {
```
mboudreau: should we make argmapping abstracted/required for instantiating super class?

src/server.ts
```diff
@@ -145,134 +67,11 @@ export class Server extends events.EventEmitter {
...
-			config.agentOptions.rejectUnauthorized = false;
-			config.agentOptions.requestCert = false;
-			config.agentOptions.agent = false;
-		}
-
-		http(config, (err: any, res: any) => (!err && res.statusCode === 200) ? deferred.resolve() : deferred.reject(`HTTP Error: '${JSON.stringify(err ? err : res)}'`));
-
-		return deferred.promise;
+	protected spawnBinary(): ChildProcess {
+		return pactUtil.spawnBinary(`${pact.mockServicePath} service`, this.options, this.__argMapping);
```
mboudreau: I feel like this could be abstracted further?

README.md
```diff
@@ -52,70 +78,37 @@ To see the list commands possible with the CLI, simply ask for help `$# pact --h
...
+| `sslcert`| false |  string          | Path to a custom self-signed SSL cert file, 'ssl' option must be set to true to use this option, defaults to none |
+| `sslkey` | false |  string          | Path a custom key and self-signed SSL cert key file, 'ssl' option must be set to true to use this, defaults to none |
+| `cors`   | false |  boolean         | Allow CORS OPTION requests to be accepted, defaults to 'false' |
+| `dir`    | false |  string          | Directory to write the pact contracts relative to the current working directory, defaults to none  |
+| `spec`   | false | number           | The pact specification version to use when writing pact contracts, defaults to '1' |
+| `consumer`          | false  | string  | The name of the consumer to be written to the pact contracts, defaults to none |
+| `provider`          | false  | string  |  The name of the provider to be written to the pact contracts, defaults to none |
+| `pactFileWriteMode` | false  | string  | Control how the pact file is created. One of 'overwrite', 'update' or 'merge', defaults to 'overwrite' |
```
mboudreau: Instead of `string` should we use TS types? like `"overwrite" | "update" | "merge"` instead?

src/service.ts
```diff
@@ -0,0 +1,243 @@
+// tslint:disable:no-string-literal
+
+import path = require("path");
+import fs = require("fs");
+import events = require("events");
+import http = require("request");
+import q = require("q");
+import logger from "./logger";
+import pactUtil, {DEFAULT_ARG, SpawnArguments} from "./pact-util";
```
mboudreau: `DEFAULT_ARG` not being used.

src/service.ts
```diff
@@ -0,0 +1,243 @@
...
+import {ChildProcess} from "child_process";
+const mkdirp = require("mkdirp");
+const pact = require("@pact-foundation/pact-standalone");
```
mboudreau: `pact` not being used.

README.md
mefellows: Sure!

src/pact.ts
mefellows: Good call, I'll update the definition for the mock too.

src/server.ts
mefellows: It already is, it's defined as an abstract property in the `AbstractService` class, so if not defined the class doesn't even compile. Is that what you mean?

src/server.ts
mefellows: Arguably. I thought about a few options but nothing felt 'good'. I landed on this as I felt it was the most clear, but if you had thoughts on a nice way I'm keen to hear.

src/service.ts
mefellows: I need to fix my linter, clearly. This should be a no-brainer for automation.

src/server.ts
mboudreau: ah, that seems a bit weird as a readonly property.  I'd probably add it as an argument of the AbstractService constructor or possibly as an abstract get function.

src/server.ts
mboudreau: well, they all do exactly the same thing except for the first argument.  Maybe this is something else that can go in the AbstractService constructor?

src/service.ts
mboudreau: or, you can use a good IDE ;)

bethesque:
The options aren't actually the same. The stub has a subset.

src/server.ts
mefellows: Yeah, I thought that but felt wasn't ideal. Second opinion is always good.

I've pushed up an example, let me know what you think.

src/service.ts
mefellows: I've added to `tslint.json` actually, it will fail a compile if we have dangling variables that aren't being used.

src/service.ts
```diff
@@ -0,0 +1,246 @@
...
+		};
+	}
+
+	public readonly options: ServiceOptions;
+	protected abstract readonly __argMapping: any;
+	protected __running: boolean;
+	protected __instance: ChildProcess;
+	protected __serviceCommand: string;
+
+	constructor(command: string, options: ServiceOptions) {
```
mefellows: Let me know what you think of this param...

src/server.ts
mefellows: Fixed.

mefellows:
Have added your feedback and pushed. I think we are good to go unless you want to change the constructor options. Let me know!

bin/pact-cli.ts
```diff
@@ -26,6 +26,18 @@ cli
 	.option("--provider [providerName]", "Specify provider name for written Pact files.")
 	.action((args: any, options: any) => pact.createServer(options).start());
 
+cli
+	.command("stub")
+	.description("Creates an API stub from pact files")
+	.option("-p, --port [n]", "Port on which to run the service. Default is random.", cli.INT)
```
mboudreau: Please change all square brackets for angle brackets as per my new feature breach.

src/pact.ts
```diff
@@ -56,6 +60,48 @@ export class Pact {
...
+		if(this.__stubs.length === 0) {
+			return q(this.__stubs);
+		}
+
+		logger.info("Removing all Pact stubs.");
+		return q.all<Stub>(_.map(this.__stubs, (stub:Stub) => stub.delete() as PromiseLike<Stub>));
+	}
+
+	// Remove all the servers and stubs
+	public removeAll(): q.Promise<any> {
```
mboudreau: Should the return type be a promise with AbstractService?

bin/pact-cli.ts
mefellows: Done.

src/pact.ts
mefellows: Arguably, but the compiler didn't agree with me. I thought given both Stub and Server were subtypes this would work - but alas, no. 

I could create an interface so that it should work given its structural type, but that felt even worse.

Ideas?

mboudreau:
Ah, yeah, q works with generics, so you have to do q.all<AbstractService>

mefellows:
Yeah. I mean, I'm not a TypeScript expert but I have been working with generics for a long time.

As in, this code doesn't compile:

```
	// Remove all the servers and stubs
	public removeAll(): q.Promise<AbstractService> {
		return q.all<AbstractService>([this.removeAllStubs(), this.removeAllServers()]);
	}
```

Error:
```
src/pact.ts(102,33): error TS2345: Argument of type '(Promise<Server[]> | Promise<Stub[]>)[]' is not assignable to parameter of type 'IWhenable<IWhenable<AbstractService>[]>'.
  Type '(Promise<Server[]> | Promise<Stub[]>)[]' is not assignable to type 'PromiseLike<IWhenable<AbstractService>[]>'.
    Property 'then' is missing in type '(Promise<Server[]> | Promise<Stub[]>)[]'.
```

I think the problem is the shape of the Promise, and not the type it contains. I can have a bit more of a play, could just be a brain fart.

mefellows:
After further investigation, I think there may be an issue with the type itself as it doesn't seem to line up with the [docs](https://www.npmjs.com/package/q#combination) or my understanding.

I could only find the following:

* https://github.com/DefinitelyTyped/DefinitelyTyped/issues/17878

...but what I find curious is the signature of `q.all()` as defined by the types package:

```
export function all<T>(promises: IWhenable<Array<IWhenable<T>>>): Promise<T[]>;
```

I'd expect the first argument to be an array, not a Promise of an Array. i.e. I'd expect the signature to be:

```
export function all<T>(promises: Array<IWhenable<T>>): Promise<T[]>;
```

So this is why it will never compile, even though it works fine at runtime. 

Any suggestions, or should we just move this in? One thing is that given the current type is any, narrowing the type after we bring this in wouldn't be a breaking change, so it's not irreversible.

PR #68, opened 2018-01-03
docs: Correct required column to match the current behaviour

TimothyJones:
The `Required?` column now matches the `Description`, and the current behaviour.


mboudreau:
Good find, documentation and some of the code appears to be wrong in terms of what is required or not.  I'm going to abandon this PR as I'll update the code and docs in a combined commit.  Thank you for bringing this to our attention.  Cheers.

mboudreau:
@TimothyJones I've updated the docs and the code, new version out with the changes, 6.7.1 that's currently building/deploying.  Cheers.

PR #76, opened 2018-03-14
Merging pact standalone into pact-node to make upgrading easier

mboudreau:
Here it is in all it's glory.  Took me a while to achieve, but it's finally done. I've tested it thoroughly internally to make sure everything is still kosher, but I haven't tried it "in the wild" since it's based on a postinstall script to download the appropriate binary.  I might just try to do an RC build or something so I can test it out with another project.

@bethesque while I was implementing the new standalone, I updated it to the latest version of the ruby portable one and noticed that while I was running tests, it would create a few folders on the root of my git repo, folders like `log`, `reports` and `tmp`.  I noticed that these are actually default references within the pact ruby files.  Are these new?  I tried to figure out how to set them to a tmp directory, but I was getting some issues doing so.  Also, it is possible not to have it pollute my current working directory with all these folders?  Maybe consolidate them to a single "pact" folder of some kind which will make it easier for people to add to .gitignore


mefellows:
Looks ace @mboudreau, great stuff! I'll take a look through over the next couple of days. This is going to make our release workflow so much nicer, not to mention the reduction in size of the distribution 😲

bethesque:
@mboudreau are the folders being created during the consumer tests or the provider tests?

mboudreau:
Provider. I believe it's the verification. You can find the paths for the
directories in the configuration.rb file under the provider folder.

standalone/install.ts
```diff
@@ -0,0 +1,112 @@
+"use strict";
```
mefellows: Any reason we have the `"use strict";` here but not in other files? We should probably also consider the use of the  strict flag on the compiler also.

src/pact.spec.ts
```diff
@@ -237,27 +235,26 @@ describe("Pact Spec", () => {
...
-				pactBroker: "http://localhost",
-				pactFilesOrDirs: [path.dirname(currentDir)],
-				consumerVersion: "1.0.0"
-			};
-			return expect(pact.publishPacts(opts)).to.eventually.be.fulfilled;
-		});
-	});
-})
-;
+	/*describe.skip("Verify Pacts", () => {
```
mefellows: Let's just remove these - they are very much integration tests that belong in the other project anyhow.

mefellows: We also have tests in the integration folder for them.

standalone/install.ts
mboudreau: ah, it was originally a javascript file.  I need to remove that.  typescript doesn't need strict :P

src/pact.spec.ts
mboudreau: ah cool, will do.

mboudreau:
Appveyor is being a pain.  From my research, it seems like there's an issue with NPM above 5.4 with permissions on windows.  I might have to find an alternative to unzip the file, maybe execute a CLI command instead.  Completely ridiculous.

bethesque:
I've updated pact and pact-support to remove the tmp and reports dirs @mboudreau, but the log one would be more work, and I don't think is such a problem. Most repositories ignore the log dir anyway (and to be honest, should ignore the tmp directory too). I'll do a release of the standalone when I have a moment.

mboudreau:
@bethesque thanks.  Can we make it so that it's configurable in the CLI as an option?

bethesque:
Yes, but tbh, my first priority is getting pact-message done, so I won't put it in this release. Raise it as an issue in the pact-provider-verifier project.

bethesque:
I've released 1.30.1 of the standalone with the tmp and reports dir fixes.

mboudreau:
Finally! Stupid windows :/

mboudreau:
@mefellows want to take a last peek before I merge and deploy a new version?

mefellows:
Sure...

mefellows:
LGTM. 

If I understand correctly, you're basically now installing each OS package into its own directory to avoid the file lock problem (on Windows) and have added some methods/fields to capture this information.

Sounds reasonable, I say ship it!

mboudreau:
@mefellows @bethesque Alright, pact-standalone should be no more, latest version 6.12.0 has the new changes.  

@mefellows please update pact-js to the latest version. 

I'll create a deprecation message for pact-standalone project so people sees it and hopefully upgrades their pact-node/pact-js version to the latest.

Cheers.

mefellows:
`5.8.0` on its way out!

PR #81, opened 2018-03-26
feat(messages): add Message Pact support

mefellows:
- Adds support for invoking the pact-message binary from the
latest standalone package
- Upgrades standalone to 1.33.1

_NOTE_: I also noticed an [issue](https://github.com/palantir/tslint/issues/3644) with our linting setup, and have adjusted the PR to accommodate as I was unable to run linting locally before pushing. I made the assumption that the 160 chars character was what was originally intended, and modified the source to suit. Interestingly, this meant some other lint issues weren't found, which I fixed in commit 9c0637ccbf2d02b461a77f71b975ace799a18ced.


mboudreau:
you should always run `npm test` before committing/pushing ;)

mboudreau:
Can you merge my latest changes to this PR?  I think I've fixed some more windows issues with appveyor (yay windows file locking for no reason!)

mefellows:
Weirdly enough, I _did_ run `npm t` before pushing, but that didn't cause any issues! It now correctly runs the linting. 

And yes, will rebase and push now.

src/server.spec.ts
```diff
@@ -173,9 +173,9 @@ describe("Server Spec", () => {
 			});
 
 			it("should start correctly with monkeypatch", () => {
-				const server = serverFactory({monkeypatch: monkeypatchFile});
-				expect(server.options.monkeypatch).to.equal(monkeypatchFile);
-				return expect(server.start()).to.eventually.be.fulfilled;
+				const s = serverFactory({monkeypatch: monkeypatchFile});
```
mefellows: Could have disabled a linting rule here (`no-shadow`) but felt this was less evil.

test/integration/data-utils.ts
```diff
@@ -14,7 +14,7 @@ export function auth(req: express.Request, res: express.Response, next: express.
 	if (user && user.name === "foo" && user.pass === "bar") {
 		next();
 	} else {
-		res.set("WWW-Authenticate", "Basic realm=Authorization Required").sendStatus(401)
+		res.set("WWW-Authenticate", "Basic realm=Authorization Required").sendStatus(401);
```
mefellows: Somehow this wasn't being picked up?

mefellows:
All rebased and ready to go!

bin/pact-cli.ts
```diff
@@ -16,23 +16,34 @@ cli
 	.option("-h, --host <hostname>", "Host on which to bind the service. Default is localhost.")
 	.option("-l, --log <file>", "File to which to log output to.")
 	.option("-s, --ssl <boolean>", "Use a self-signed SSL cert to run the service over HTTPS. Default is false (HTTP).", cli.BOOL)
-	.option("-o, --cors <boolean>", "Support browser security in tests by responding to OPTIONS requests and adding CORS headers to mocked responses. Default is false.", cli.BOOL)
+	.option("-o, --cors <boolean>", `Support browser security in tests by responding to OPTIONS requests and adding CORS headers
```
mboudreau: Did your IDE wrap all of these strings?  You do know this is going to break many of them at runtime?

src/message.ts
```diff
@@ -0,0 +1,93 @@
...
+import pactUtil, {DEFAULT_ARG, SpawnArguments} from "./pact-util";
+import pactStandalone from "./pact-standalone";
+import path = require("path");
+const mkdirp = require("mkdirp");
+const checkTypes = require("check-types");
+
+export class Message {
+	public readonly options: MessageOptions;
+	private readonly __argMapping = {
+		"content": DEFAULT_ARG,
```
mboudreau: I'm not sure that we can do 2 default args?  It's also a fairly bad CLI practice.

src/server.spec.ts
mboudreau: :+1:

src/message.ts
mefellows: It seems to work, at least, the integration test does.

bin/pact-cli.ts
mefellows: I wrapped them by hand on a logical break point (to address the 160 chars limit).

What do you mean breaks at runtime? I tested by running the cli locally, e.g. here is an output for mock:

```sh
matt ~/development/public/pact-node λ node ./bin/pact-cli.js mock  --help

   pact 6.12.2 - Pact CLI tool

   USAGE

     pact mock

   OPTIONS

     -p, --port <n>                      Port on which to run the service. Default is random.                                                                                    optional
     -h, --host <hostname>               Host on which to bind the service. Default is localhost.                                                                                optional
     -l, --log <file>                    File to which to log output to.                                                                                                         optional
     -s, --ssl <boolean>                 Use a self-signed SSL cert to run the service over HTTPS. Default is false (HTTP).                                                      optional
     -o, --cors <boolean>                Support browser security in tests by responding to OPTIONS requests and adding CORS headers to mocked responses. Default is false.      optional
     -d, --pact-dir <directory>          Directory to which the pacts will be written. Default is cwd.                                                                           optional
     -i, --pact-version <n>              The Pact specification version to use when writing the Pact files. Default is 1.                                                        optional
     -w, --pact-file-write-mode <m>      Controls how pact files are written to disk. One of 'overwrite', 'update', 'merge'                                                      optional
     --consumer <consumerName>           Specify consumer name for written Pact files.                                                                                           optional
     --provider <providerName>           Specify provider name for written Pact files.                                                                                           optional
     --monkeypatch <file>                Absolute path to a Ruby file that will monkeypatch the underlying Pact mock.                                                            optional

   GLOBAL OPTIONS

     -h, --help         Display help
     -V, --version      Display version
     --no-color         Disable colors
     --quiet            Quiet mode - only displays warn and error messages
     -v, --verbose      Verbose mode - will also output debug messages
```

Looks ok?

mboudreau: Ah, it's because it's using ` and not ".  The latter would make it break since string cannot be multiline without having a + between them.

mefellows: Hmm, no, you're right. This is fixed now.

src/message.ts
mboudreau: But why both?  Are they both mandatory?

bin/pact-cli.ts
mefellows: ... for posterity, the `` ` `` works, but it actually leaves the space in (so formatting is shit). Fixed with `"` .

src/message.ts
mefellows: Yep, they're both mandatory.

mboudreau: alright, but I'm not a fan :/

mefellows:
There was a weird build failure before - just needed to re-run it. Thanks for the approval - merging!

mboudreau:
You going to create a new release?

mefellows:
Yep, should be done by now

PR #87, opened 2018-04-12
begging for money on postinstall with opencollective link

mboudreau:


mboudreau:
That is a very good idea. Maybe I'll have them go through donations.pact.io
and add some query about which tech it's coming from.

mboudreau:
@mefellows added google analytics to create a redirect from donate.pact.io to opencollective, and it works! :D

mefellows:
Let's doe this!

mboudreau:
Actually, wait, forgot to add "node" to the query so we can track the
origin.

mboudreau:
Alright, while I was in there, also added google analytics tracking for downloads that should give us a better idea of how many times pact-node is actually installed separate from npm stats.

standalone/install.ts
```diff
@@ -14,9 +14,29 @@ function download(data: Data): Promise<Data> {
...
 		}
-
 		console.log(chalk.yellow(`Downloading Pact Standalone Binary v${PACT_STANDALONE_VERSION} for platform ${data.platform} from ${data.url}`));
 
+		// Track downloads through Google Analytics unless testing or don't want to be tracked
+		if (!process.env.DO_NOT_TRACK) {
+			console.log(chalk.gray("Please note: we are tracking this download anonymously to gather important usage statistics. " +
+			"To disable tracking, set 'DO_NOT_TRACK=true' as an environment variable."));
+			const CI = ["CI", "CONTINUOUS_INTEGRATION"].some((key) => process.env[key] !== undefined);
+			request.post({
```
mefellows: I think this is the issue with #91, it should fail gently with perhaps a note at `DEBUG` level that it didn't send any metrics.

PR #89, opened 2018-04-30
feat(env): expose pact standalone version and base url as env variables

uldissturms:
In our organization we are behind a firewall and cannot install dependencies form external sources.
This pull request is to introduce capability to specify custom `PACT_STANDALONE_BASE_URL` so that node-pact binaries can be downloaded from internal artifact repository. At the same time introducing setting `PACT_STANDALONE_VERSION` from env variables in case the need arises.


uldissturms:
AppVeyor builds failed for node versions 5 and below.
Seems unrelated to the changes in this PR though. Travis happy for Linux & Mac.

`C:\Users\appveyor\AppData\Roaming\npm\node_modules\npm\bin\npm-cli.js:79
698      let notifier = require('update-notifier')({pkg})`

Can anyone with the right perms kick the build again?

Issue #88, opened 2018-04-24
Behind a firewall pact-node fails to install


vrish88:
We have a CI environment behind a firewall, so we fail to install pact-node because it tries to download the server binaries from [github](https://github.com/pact-foundation/pact-node/blob/master/standalone/install.ts#L124).

Is there a way to handle an install behind a firewall? I can think of a couple ways we could get around this. There may be a way to specify a location to download the binaries from (like an on-prem artifactory server or something). Alternatively, we could download the binaries and put them into our codebase and (somehow) tell pact-node where they are on an install.

mboudreau:
Well, first off, I would say that this issue wouldn't be only for pact, but
any node modules as they would have to be downloaded from npm regardless.

The best option for you, to reduce maintenance costs, would be to whitelist
github releases domain on your network.

The other option would be to do `npm install @pact-foundation/pact-node` on
your work computer, then commit the resulting directory
`node_modules/@pact-foundation/pact-node`. However, this is going to be a
pain in the neck moving forward of you want to update the version and have
to manually change it, update and commit the resulting files.

I would argue that this workaround makes the CI firewall completely useless
since you can still have 3rd party binaries in git, circumventing any
security the firewall provides. I would try to change that policy
internally since this isn't conducive for quick innovation.

mefellows:
What we're doing is pretty non standard though Michel, so I can see why this is coming up. It's not the first time time too.

Is it that the GitHub URL isn't downloadable (can you download that URL manuall). It might just he that certain environment vars need to be exposed and we need to update our docs.

E.g. `HTTPS_PROXY`

vrish88:
@mboudreau right now we've got holes poked in the firewall for npm/yarn repositories so we are able to download js dependencies. If we had control over the firewall we totally would whitelist the github releases domain .

We are able to download the npm dependencies just fine in our CI env. We can download the resources from github on our local development machines. Unfortunately it's our CI environment that blocks github. For our current situation we have the ability to upload resources elsewhere. So it would be janky and manual, but we could download them locally, and then upload them somewhere our CI could get them.

To enable that it could be as simple as replacing https://github.com/pact-foundation/pact-node/blob/357d615f687708f55ba57d71dd4f0981d367ca81/standalone/install.ts#L123-L130 with something like

```javascript
[...]

PR #92, opened 2018-05-16
feature(checksum): adding checksum validation

mboudreau:
Doing validation checksum of pact standalone binaries to make sure they are correct, allowing the use to specify a different URL or local path to the location of the binaries without compromising on the security.  This will fix this issue: https://github.com/pact-foundation/pact-node/issues/88

I still need to document the new environment variable available `PACT_BINARY_LOCATION`, which I might add a way to add it to package.json as a config as well.


standalone/install.ts
```diff
@@ -60,31 +68,46 @@ function download(data: Data): Promise<Data> {
 					cd: `download-node-${data.platform}-${isCI ? "ci" : "user"}`
 				}
 			})
-				// Ignore all errors
-				.on("error", () => {});
+			// Ignore all errors
```
mefellows: 👍

standalone/install.ts
```diff
@@ -1,20 +1,28 @@
...
 const path = require("path");
 const fs = require("fs");
+const urljoin = require("url-join");
 const decompress = require("decompress");
 const tar = require("tar");
 const chalk = require("chalk");
+const sumchecker = require("sumchecker");
+
+export const PACT_STANDALONE_VERSION = "1.44.0";
+export const PACT_BINARY_LOCATION = process.env.PACT_BINARY_LOCATION ||
```
mefellows: I thought we were going with the entry in `package.json` as per #88?

standalone/install.ts
```diff
@@ -60,31 +68,46 @@ function download(data: Data): Promise<Data> {
...
+			// Or else it might be a local file, try to copy it over to the correct directory
+			fs.createReadStream(data.binaryDownloadPath)
+				.on("error", (e: string) => reject(`Error reading the file at '${data.binaryDownloadPath}': ${e}`))
+				.pipe(
+					fs.createWriteStream(data.filepath)
+						.on("error", (e: string) => reject(`Error writing the file to '${data.filepath}': ${e}`))
+						.on("close", () => resolve(data))
+				);
+		} else {
+			reject(`Could not get binary from locale ${data.binaryDownloadPath} as it's not a URL and does not exist at the path specified.`);
```
mefellows: "Locale" or "local"?

standalone/install.ts
mboudreau: Yeah, still need to figure that out as I'm not sure how to get the package.json of the containing project...

standalone/install.ts
mboudreau: local :|

standalone/install.ts
```diff
@@ -132,38 +167,84 @@ function extract(data: Data): Promise<void> {
...
+		platformFolderPath: path.resolve(__dirname, entry.folderName)
+	});
+}
+
+function join(...paths: string[]): string {
+	return HTTP_REGEX.test(paths[0]) ? urljoin(...paths) : path.join(...paths);
+}
+
+const checksumSuffix = ".checksum";
+const BINARIES: BinaryEntry[] = [
```
mefellows: Nice, i think this makes the function a bit more readable.

standalone/install.ts
mboudreau: Not sure this is possible.  Seems that it only works if the thing runs through NPM itself, like (`npm run whatever`).  I might just stick to environment variables cause I think it just might be simpler for now.  We can always adapt it later.

mboudreau:
na-na nah! super nerd!

mboudreau:
Hm, apparently, only windows, pact-node isn't dispatching the 'started' event when starting the mock service.  Weird.  I wonder what changed.

mboudreau:
Issue seems to be coming from the stub service.  I think there was a change recently that makes the stub service fail when setting a windows path to a local file.  Will need to fix the binary issue before this change can go out sadly.

PR #103, opened 2018-07-19
use the OS tmp directory to downloading and run binaries

mboudreau:
Try to prevent windows path length limitations and errors


src/pact-util.ts
```diff
@@ -10,8 +11,8 @@ const checkTypes = require("check-types");
 export const DEFAULT_ARG = "DEFAULT";
 
 export class PactUtil {
-	public get cwd():string {
-		return path.resolve(__dirname, "..");
+	public get binaryBasePath():string {
+		return path.join(os.tmpdir(), "pact");
```
mefellows: On Mac/Linux, won't this result in the pact binary mysteriously "disappearing"? Files in `$TMPDIR` will be cleaned up occasionally by the OS on a Mac (default every 3 days) and on a reboot on Linux.

mboudreau:
Yeah, Beth brought that up today. Not sure that windows does the same
thing, but we could have it be different for each os.

mefellows:
According to my research, it seems Windows never cleans up its temp directory. But I'm not sure this will help - the default TEMP directory is different across Windows versions anyway, and can actually be quite long. E.g. Vista + beyonds temp dir is `C:\Users\<User Name>\AppData\Local\Temp`.

Is the custom download location not a satisfactory alternative to this?

TimothyJones:
Closing this for now, as it seems like the solution to the windows path length issue might lie elsewhere. Feel free to reopen.

Issue #83, opened 2018-04-03
Feature Request: Create binstubs for pact-standalone binaries


mefellows:
AS a user who wants to integrate to the latest Pact CLI features
IT would be useful to be able to easily reference the Pact binaries (e.g. `pact`, `pact-broker` `pact-message` etc.) 
SO that I can use them without having to write JS wrappers, or wait for them to be included in the JS library

**Example use case**

Prior to deployment, we often run the `can-i-deploy` tool to check if it's safe to release either a consumer or a provider. Bash is an appropriate tool for this. It's hard to find the CLI tools if you don't know what you're doing, but it is possible:

package.json:
```
"can-i-deploy:consumer": "$(find ./node_modules -name pact-broker | grep -e 'bin/pact-broker$' | head -n 1) can-i-deploy --pacticipant SNSPactEventConsumer --latest --broker-base-url https://test.pact.dius.com.au --broker-username xxx --broker-password yyy",
```

Typically, I just install the CLI tools into my Docker container and add them to the path. But it would be nice if I could just go:

package.json:
```
can-i-deploy:consumer": "pact-broker can-i-deploy --pacticipant SNSPactEventConsumer --latest --broker-base-url https://test.pact.dius.com.au --broker-username xxx --broker-password yy
```

I imagine the `./node_modules/.bin` dir might look like:

```
ls ./node_modules/.bin
pact_broker -> ./node_modules/@pact-foundation/pact-node/standalone/darwin-1.33.1/bin/pact-broker
pact -> ./node_modules/@pact-foundation/pact-node/standalone/darwin-1.33.1/bin/pact
...

```

mefellows:
Did you want to have a stab at a PR for it?

amr-labib:
Yes, I could try, it would be my first github PR though. I am reading the contribution guidelines now.

mefellows:
That would be great, thanks!

amr-labib:
I am working on getting this done but I found that we will run into a conflict with one of the binaries. When I look at one of my projects with `pact-node` installed, I find a symlink in `./node_modules/.bin`  defined as:

`pact -> '../@pact-foundation/pact-node/bin/pact-cli.js'`

This will conflict with the pact binary in the standalone package.

What would be recommended here? Should I just ignore that one?

mefellows:
[...]

PR #106, opened 2018-08-13
Add can-i-deploy to API and CLI tool

TimothyJones:
This PR adds the `can-i-deploy` to the API and CLI tools.

Since it's my first substantial change to pact-node, I'd like to highlight some discussion points:

* Because the order of arguments to `can-i-deploy` matters, but the previous way of handing options off to `PactUtil` doesn't let you specify order of arguments (or even the same argument more than once), I've added a helper in `CanDeploy` to take the options object and convert it into an array of options. Possibly this helper should be in `PactUtil` instead? I'm torn - it's specific to `can-i-deploy`, but it's not really the responsibility of the `CanDeploy` object.

* All the functions / classes here are named `CanDeploy`. I'm not sure if I should have used `CanIDeploy` instead?

* The first commit on this branch also bumped the version number - but I wasn't sure if I should have removed the bump before making the PR?

*  We have a small challenge with the arguments to the CLI -  the broker binary CLI accepts `can-i-deploy --pacticipant <p1> --pacticipant <p2> --version <v2>` - ie, specifying pacticipants without a given version. However, because of the way the CLI library we're using for node works, we can't specify the same argument multiple times. I've changed the format to `--pacticipants "<p1>,<p2>" --versions "<v1>,<v2>"`.  At the moment, I'm rejecting arguments where the lengths of the lists don't match - which enforces that all pacticipants must have a version supplied. Although this is recommended as good practice anyway, we may not want to enforce it. I don't like having a difference between what is possible to specify using the node cli and the binary cli, but also I want the node cli arguments to be easy to use correctly.

* This PR doesn't add the documentation yet - I figured I'd write it after we discuss the above.

Comments and review very welcome!

### DIFF
--- a/bin/pact-cli.spec.ts
+++ b/bin/pact-cli.spec.ts
@@ -125,6 +125,35 @@ describe("Pact CLI Spec", () => {
 			});
 		});
 	});
+
+	describe("can-i-deploy Command", () => {
+		it("should display help", () => {
+			const p = CLI.runSync(["can-i-deploy", "--help"]).then((cp) => cp.stdout);
+			return q.all([
+				expect(p).to.eventually.contain("USAGE"),
+				expect(p).to.eventually.contain("pact can-i-deploy")
+			]);
+		});
+
+		it("should fail if missing arguments", () => {
+			return expect(CLI.runSync(["can-i-deploy"]).then((cp) => cp.stderr)).to.eventually.contain("Error");
+		});
+
+		context("with broker mock", () => {
+			const PORT = 9123;
+			const brokerBaseUrl = `http://localhost:${PORT}`;
+			let server: http.Server;
+
+			before(() => brokerMock(PORT).then((s) => server = s));
+			after(() => server.close());
+
+			it("should work pointing to fake broker", () => {
+				const p = CLI.runSync(["can-i-deploy", "--pacticipants", "pacticipant1", "--versions", "1.0.0", "--pact-broker", brokerBaseUrl])
+					.then((cp) => cp.stdout);
+				return expect(p).to.eventually.be.fulfilled;
+			});
+		});
+	});
 });
 
 class CLI {
@@ -159,7 +188,7 @@ class CLI {
 		return this.run(args)
 			.tap((cp) => {
 				if ((cp.process as any).exitCode === null) {
-					// console.log("check when exiting");
+					//console.log("check when exiting");
 					const deferred = q.defer<CLI>();
 					cp.process.once("exit", () => deferred.resolve());
 					return deferred.promise;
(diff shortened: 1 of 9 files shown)


mefellows:
I hear your pains/concerns Tim. I was going to add this feature in a while back but ran into the same inertia.

I'll take a look on Thursday when I'm not on client site.

Another consideration is #83 - is it worth doing or just exposing it out via binstubs (and therefore just providing a convenient installer for the CLI tools).

I still think there is value in exposing via a JS interface though.

TimothyJones:
Oh, I hadn’t seen that issue, thanks! It looks like it would largely obviate the need for pact-cli.js?

I did wonder if the JS api should ask for the json response from can-I-deploy, parse it and return?

Until I made this PR, I didn’t even realise that the cli existed and could be called from scripts in package.json (which is awesome). Is this documented somewhere?

mefellows:
> Oh, I hadn’t seen that issue, thanks! It looks like it would largely obviate the need for pact-cli.js?

For me, yes. I don't see the need to wrap the Ruby CLI with a JS CLI. 

> I did wonder if the JS api should ask for the json response from can-I-deploy, parse it and return?

I think so, it enriches the interface significantly. I'd like to do the same with the provider verifier too if time permits.

> Until I made this PR, I didn’t even realise that the cli existed and could be called from scripts in package.json (which is awesome). Is this documented somewhere?

There is a small section under [usage](https://github.com/pact-foundation/pact-node#usage) but largely not visible. 

I'll hopefully get to put some more thinking into this later in the week, but by and large I think we would like to see this in the library. The whole point of Pact Node is to simplify operations with the Pact CLI suite. If all it does is just package the binaries it's falling short of that goal.

mefellows:
> Cool. I’m a little confused about the difference between pact-node and pact-js, too.

You're not the first to say this. Long term, they may merge.

But for now think of it this way:

* `pact-js`: The client facing interface used in test suites. 9 times out of 10, it is the package you want to use.
* `pact-node`: generic wrapper of the Pact CLI interfaces callable from JS (it started as just being an event emitter for the mock server but evolved over time). Think of it more as a "library".

> On the packaging of the ruby libraries, it is really nice to be able to write scripts in package json that directly call the cli- but solving that other issue would mean this works without the wrapper.

Exactly. It also means when new features are added to the CLI they are easily added to this package.

lirantal:
Would be great to see this land as a standalone CLI binary for `canideploy` instead of having to bash hack the node_modules/ dir to find it.

About the confusion between pact-js and pact-node, I was under the impression that a more distinctive guideline would be that pact-js is for frontend libs where you don't require to have the pact mock service started by the test, where-as with pact-node you get that bundled in already and don't need to also worry about bringing it up.

lirantal:
I'd actually like to extend this issue and ask if it will be able to expose the `can-i-deploy` functionality as a library that I can call from my own Node.js tools/project ?

mboudreau:
I merged it, fixed a few things and released it as version 6.21.0.  Cheers.

lirantal:
Great job @TimothyJones!

Issue #104, opened 2018-08-10
Unable to run npm install on a fresh checkout


TimothyJones:
On a fresh checkout, `npm install` fails:

```
$ git clone git@github.com:pact-foundation/pact-node.git
$ cd pact-node
$ npm install
....

> @pact-foundation/pact-node@6.19.8 postinstall /Users/work/office/pact/test/pact-node
> node postinstall.js

internal/modules/cjs/loader.js:583
    throw err;
    ^

Error: Cannot find module './standalone/install'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
    at Function.Module._load (internal/modules/cjs/loader.js:507:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/Users/work/office/pact/test/pact-node/postinstall.js:1:79)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @pact-foundation/pact-node@6.19.8 postinstall: `node postinstall.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the @pact-foundation/pact-node@6.19.8 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/work/.npm/_logs/2018-08-10T04_15_32_144Z-debug.log
```

doing `npm run build` (which also fails, because the dependencies are missing), will generate the module, but then `npm install` fails differently:

```
$ npm install
.... 

Finished downloading binary to /Users/work/office/pact/test/pact-node/standalone/pact-1.54.1-osx.tar.gz
/Users/work/office/pact/test/pact-node/standalone/install.js:263
    throw new Error(chalk.red("Unhandled Promise Rejection: " + e));
    ^

Error: Unhandled Promise Rejection: Postinstalled Failed Unexpectedly: Checksum file missing from standalone directory. Aborting.
[...]

PR #108, opened 2018-08-13
scripts(package.json): Fix broken npm install command on a fresh checkout

TimothyJones:
This fixes issue #104, by ensuring that the typescript is built and the checksums are downloaded.

The install worked in CI, because the test scripts do this too, and install was run with `--ignore-scripts`. It's possible that we should remove `--ignore-scripts` from CI, to more quickly surface this sort of thing if it happens again? I'm not sure why `--ignore-scripts` is run in CI, so I haven't removed it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "typescript": "2.7.2"
   },
   "scripts": {
-    "postinstall": "node postinstall.js",
+    "postinstall": "npm run build && npm run download-checksums && node postinstall.js",
     "install:all": "npm run install:osx && npm run install:win && npm run install:linux32 && npm run install:linux64 && npm run install:current",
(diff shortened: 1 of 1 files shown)


mefellows:
Hmm how would this work after the package is installed after distribution? I don't think we can assume TS is installed on the target.

TimothyJones:
Oh, hmm. Perhaps I don't know enough about the npm package installation process. A quicker fix might be to update the documentation to say to use `--ignore-scripts`?

I also found the tests very brittle on my local install (so brittle that I disabled everything that wasn't the spec I was working on). I don't know if that's related.

TimothyJones:
I created #111 instead

PR #131, opened 2018-10-23
build(deps): Fix GitHub warning by updating request from 2.83.0 to 2.87.0

TimothyJones:
# The problem

Currently a repository including `pact-node` gets the following warnings in the GitHub gui:

screen shot 2018-10-23 at 12 27 12 pm

screen shot 2018-10-23 at 12 26 50 pm

For `request@2.83.0`,  the problem library `cryptiles` was brought in by `hawk`:

```
$ npm ls cryptiles
@pact-foundation/pact-node@6.19.11 /Users/work/office/pact/pact-node
└─┬ request@2.83.0
  └─┬ hawk@6.0.2
    └── cryptiles@3.1.2
```

# The solution

This PR upgrades `request`. Since later versions of request no longer depend on `hawk` or `cryptiles`, this will remove the warning in future releases.

```
$ npm ls cryptiles
@pact-foundation/pact-node@6.19.11 /Users/work/office/pact/pact-node
└── (empty)
```

*Aside:* The latest version of `request` is actually `2.88.0` (as suggested by [this dependabot PR](https://github.com/pact-foundation/pact-node/pull/130)). However, the [travis build for 2.88.0](https://travis-ci.org/request/request/jobs/414594405) fails with lots of test errors, so I don't think we should use it. This is also further evidence that dependabot isn't very useful for us (at least in this configuration).

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "decompress": "4.2.0",
     "mkdirp": "0.5.1",
     "q": "1.5.1",
-    "request": "2.83.0",
+    "request": "2.87.0",
     "rimraf": "2.6.2",
     "sumchecker": "^2.0.2",
     "tar": "4.4.0",


TimothyJones:
Closing this, as:

1) the travis build for `request@2.88.0` is [broken with many test failures](https://travis-ci.org/request/request/jobs/414594405) 
2) Request is upgraded to a stable version in #131

TimothyJones:
Oops, closed the wrong one. This is the good one! Sorry for the spam.

mboudreau:
Looks good :)

Do you know how to do a release?

TimothyJones:
I don't! But happy to do it if you want to point me to the right place.

mboudreau:
Actually, I need to update the core library as well, so I'll do it in a sec :)

mboudreau:
Version 6.19.12 is going out now with the security fix, then release 6.20.0 as well with the binary update :)

PR #135, opened 2018-11-25
fix(verification): fail verification if empty [] found

### LINKED ISSUE #134: Message Provider Validation against Pact Broker Wrongly Succeeds
I have a strange behavior during message provider verification.

I successfully published a message pact to my test pact broker instance (v 2.26.0, hosted by you, if it helps ;)). When viewing the message pact in the pact broker, it says `Note: this contract could not be parsed to a v1 or v2 Pact, showing raw content instead.` and only shows the JSON.

When I run my [provider test](https://github.com/thombergs/code-examples/blob/master/pact/pact-node-messages/src/provider/hero-event-producer.spec.js), **it succeeds, even though my [message provider](https://github.com/thombergs/code-examples/blob/master/pact/pact-node-messages/src/provider/hero-event-producer.js) returns an invalid message**.

**When I run the provider test against a local pact file instead, it fails, as I would expect.**

Maybe the verifier doesn't find the pact on the broker (because the broker does not support message contracts) and since it hasn't found a pact, it just succeeds? If this is the case:
* I would expect the verifier to fail if it doesn't find any pacts to test against.
* when will the pact broker support message contracts? ;)

mefellows:
In some edge cases when following links in the broker, there are no links
to pact files in the leaf nodes. The classic case is for the 'latest' tag (not a real tag).

This change ensures the verification process fails (instead of misleadingly continues/passes):

Fixes #134

### DIFF
--- a/src/verifier.ts
+++ b/src/verifier.ts
@@ -148,6 +148,11 @@ export class Verifier {
 				}
 				return uris;
 			})
+			.tap((pacts) => {
+				if (_.isEmpty(pacts)) {
+					throw new Error (`Unable to find pacts for given provider '${this.options.provider}' and tags '${this.options.tags}'`);
+				}
+			})
 			.then((data: string[]): PromiseLike<string> => {
 				const deferred = q.defer<string>();
 				this.options.pactUrls = data;


mefellows:
Moved the failure logic to the verification process and not the broker.

mefellows:
Closing, superceded by #147

PR #138, opened 2019-01-29
fix(postinstall): Proxy support for downloading binaries

dsieradzki:
After setup proxy in npm, pact-node package is downloaded correctly but postinstall script is not able to download standalone binaries behind corporate proxy.

### DIFF
--- a/standalone/install.ts
+++ b/standalone/install.ts
@@ -129,6 +129,7 @@ function download(data: Data): Promise<Data> {
 			const isCI = CIs.some((key) => process.env[key] !== undefined);
 			request.post({
 				url: "https://www.google-analytics.com/collect",
+				proxy: process.env.npm_config_https_proxy || process.env.npm_config_proxy || undefined,
 				form: {
 					v: 1,
 					tid: "UA-117778936-1", // Tracking ID / Property ID.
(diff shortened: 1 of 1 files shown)


mefellows:
Thanks @dsieradzki, we'll take a look at this.

mboudreau:
@dsieradzki We haven't tried this yet since most of use are not using an HTTP proxy that we need to setup locally (it's done at the network level, much easier).  Did you try just using the npm proxy config? it doesn't work?  I was trying to find out if there's a way to make all node http requests go through a proxy, but I can't seem to find anything.

dsieradzki:
I have npm proxy setup and packages are downloaded. This part work well. Npm probably use proxy settings only for internal http requests (to get package itself),  while the problem is postinstall script,  because this create own "context" of http request (using "request" lib) without proxy.

mboudreau:
Yep, I think you're correct.  Have you tried updating the postinstall.js script with this new code to see if it works? Ie. does it get the npm config for the proxy correctly?

mboudreau:
I've confirmed that it does in fact work, however, instead of repeating it in 2 different locations in the postinstall, I've altered the request default so you only need to set it once and it'll affect all subsequent http/https calls.  I'll close this down as I've done it on my own branch and will release in a minute with version `6.21.3`

mboudreau:
@dsieradzki can you confirm that you've tried it and it works?

Issue #142, opened 2019-02-21
TRACKING SHOULD BE OPT-IN


chris-fran:
I just realized this library [tracks you by default](https://github.com/pact-foundation/pact-node/blob/master/README.md#do-not-track). This is not OK **and I am very disappointed**, specially since it's not documented in the readme of the MAIN library which is https://github.com/pact-foundation/pact-js.

>To respect your privacy, anyone can turn it off by simply adding a 'do not track' flag within their package.json file:

The fact that you add this line doesn't respect my privacy any more because I didn't know this was a thing in the first place. 

Furthermore, I'm not even sure this library complies with GDPR as I remember something where they mention that to collect user data you must explicitly ask them to opt-in to it and must NOT opt-in by default. Here's [the link](https://www.gdpreu.org/the-regulation/key-concepts/consent/), search for the subtitle "Clear Affirmative [Action]"

mboudreau:
We're responding here mostly: https://github.com/pact-foundation/pact-js/issues/272

mboudreau:
Closing this as we've updated our documentation on the pact-js and clarified that we are only tracking downloads of the binary only, no personally identifiable information is stored, hence we're not in breach of GDPR or any other privacy related laws.

PR #143, opened 2019-02-22
chore: anonymise IP in tracking call to GA. See https://github.com/pa…

mefellows:
See #142 for background and the following doc: https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#anonymizeIp

See https://www.blastam.com/blog/5-actionable-steps-gdpr-compliance-google-analytics for some info on GDPR, specifically:

> The definition of personal data is expanded and clarified to include IP addresses, cookie identifiers, and GPS locations.

We don't need the IP address specifically, so I suggest we remove it.

### DIFF
--- a/standalone/install.ts
+++ b/standalone/install.ts
@@ -140,7 +140,8 @@ function download(data: Data): Promise<Data> {
 					av: require("../package.json").version, // App version.
 					aid: "pact-node", // App Id.
 					aiid: `standalone-${PACT_STANDALONE_VERSION}`, // App Installer Id.
-					cd: `download-node-${data.platform}-${isCI ? "ci" : "user"}`
+					cd: `download-node-${data.platform}-${isCI ? "ci" : "user"}`,
+					aip: true, // Anonymise IP address
 				}
 			}).on("error", () => {
 			}); // Ignore all errors


mboudreau:
We don't want region approximation then?

mefellows:
Well, yes that would be preferable, but the relative data is pretty consistent with the Analytics for our docs website so we're probably not losing out on much. Some googling does seem to indicate this is a grey area, which I'd prefer to err on the right side of. Alternatively, we make it opt in (at which point we'll get basically no data).

I don't want this to be a knee-jerk reaction to the issue either, I do however, want to make sure we're seen to be taking this issue seriously.

What do you think?

TimothyJones:
Are the npm statistics really that bad?

TimothyJones:
(To be clear, I am in favour of this PR)

mefellows:
They definitely under report from what we've seen. But it does appear to be better of late (raw numbers closer). It doesn't give us stats if which versions are being used, where our users are, nor the operating environment.

I think we'll still get some indication of location, it just won't be as accurate. Country is good enough

mboudreau:
Alright, let's get it in then.

mboudreau:
version 6.21.5 is going out now.  @mefellows please update pact-js

mefellows:
Thanks Michel.

PR #147, opened 2019-03-06
feat(verify): support broker bearer token

mefellows:
- Upgrade standalone to 1.64.0
- The CLI verifier has all of the logic needed to fetch pacts
  from the Broker. This change removes the use of the Broker
  class to fetch Pacts during verification, delegating to the underlying binary
- Broker class removed completely.
- Adds support for bearer token authentication mode in the Pact Broker


README.md
```diff
@@ -251,8 +253,7 @@ pact.verifyPacts({
 | Parameter                   | Required? | Type    | Description                                                                                                |
 | --------------------------- | --------- | ------- | ---------------------------------------------------------------------------------------------------------- |
 | `providerBaseUrl`           | true      | string  | Running API provider host endpoint.                                                                        |
-| `pactBrokerBaseUrl`         | false     | string  | Base URL of the Pact Broker from which to retrieve the pacts.                                                                        |
-| `pactBrokerUrl`             | false     | string  | URL of your Pact Broker to dynamically discover relevent pacts to verify. Required if `pactUrls` not given |
```
mefellows: I've deprecated this field. So whilst it is still supported, I've removed it from the docs here and have provided a warning if it's used.

mefellows:
@mboudreau I thought about removing the Broker class entirely. We don't really document it, but technically it is still a public, usable API. If we want to keep with the idea that you can use a Node API to do certain pact things, then perhaps it's still worth keeping, otherwise I think we could make a case to remove it.

mboudreau:
@mefellows That looks pretty good.  I'd remove the broker class if I were you; even if it's exported, that's not how the API should be used as it's meant to be an internal class.

README.md
mboudreau: By removing this, we're breaking the API, hence a new version is needed.  Even with a message that's saying it's deprecated, it's still extra work for a developer to change the behaviour of their code.  I say remove all references to it everywhere, keep the deprecation message though, and increase the version to 8.0.0.

src/verifier.ts
```diff
@@ -68,12 +70,16 @@ export class Verifier {
 
 		checkTypes.assert.nonEmptyString(options.providerBaseUrl, "Must provide the providerBaseUrl argument");
 
-		if (checkTypes.emptyArray(options.pactUrls) && !options.pactBrokerUrl) {
-			throw new Error("Must provide the pactUrls argument if no brokerUrl provided");
+		if (options.pactBrokerUrl) {
```
mboudreau: I'm confused, why are we changing the name from `pactBrokerUrl` to `pactBrokerBaseUrl`?

src/broker.ts
```diff
@@ -80,6 +82,19 @@ export class Broker {
...
 			.catch(() => q.reject<IWhenable<any>>(`Unable to find pacts for given provider '${this.options.provider}' and tags '${this.options.tags}'`));
 	}
+
+	private _configureAuth()  {
+		if (this.options.username && this.options.password) {
+			return {
+				"user": this.options.username,
+				"password": this.options.password
+			};
+		} else if (this.options.token) {
```
mboudreau: just create 2 ifs instead of having an else if.  Might be good to type this function's return object like `TokenAuth | UserPassAuth`

README.md
mboudreau: I don't see why it should be deprecated.  They do exactly the same thing no?  Just the name that's different and I did that on purpose since `base` doesn't really add any value.

mboudreau:
@mefellows increment the version number to 7.1.0

src/verifier.ts
mefellows: For consistency with all of the other languages and the CLI. Alternatively, I can just leave it as is and map to the new CLI arg?

src/broker.ts
mefellows: No probs. I'm going to remove the class as suggested.

mefellows:
Thanks for the feedback. I'm going to delete the broker class, map the field back to `pactBrokerUrl` (i.e. remove the "base" bit) and bump the major version due to breaking API change (that will happen as part of release).

src/verifier.ts
mboudreau: it was already pactBrokerUrl, leave it as is.  No need to add another.

mboudreau:
@mefellows only the minor version, not the major since you're not breaking the API if you don't change that argument name.

PR #161, opened 2019-05-19
feat(binstubs): expose pact standalone binaries as alternative CLI

mefellows:
Expose the CLI as bin stubs for when the JS interface does not yet have new features, or when using JS APIs is not desirable.


TimothyJones:
This is exciting! I’ll check it out tomorrow

PR #173, opened 2019-08-04
feat(can-i-deploy): allow multiple pacticipants to be specified to CanDeploy

TimothyJones:
BREAKING CHANGE: Options for CanDeploy have changed. Now, pacticipants are specified by an array of { name: <string>, tag?: <string>, version?: <string> }, allowing more than one to be specified. If tag and version are not specified, then the latest version is used.

### DIFF
--- a/src/can-deploy.ts
+++ b/src/can-deploy.ts
@@ -12,9 +12,11 @@ export class CanDeploy {
   ): CanDeployOptions[] {
     return _.flatten(
       [_.omit(options, 'pacticipants')].concat(
-        options.pacticipants.map(({ name, tag, version }) => [
+        options.pacticipants.map(({ name, latest, version }) => [
           { name },
-          version ? { version } : { latest: tag },
+          version
+            ? { version }
+            : { latest: latest === true ? undefined : latest },
         ]),
       ),
     );
@@ -25,6 +27,7 @@ export class CanDeploy {
     name: '--pacticipant',
     version: '--version',
     latest: '--latest',
+    to: '--to',
     pactBroker: '--broker-base-url',
     pactBrokerToken: '--broker-token',
     pactBrokerUsername: '--broker-username',
@@ -113,7 +116,7 @@ export default (options: CanDeployOptions) => new CanDeploy(options);
 export interface CanDeployPacticipant {
   name: string;
   version?: string;
-  tag?: string;
+  latest?: string | boolean;
 }
 
 export interface CanDeployOptions {
@@ -124,6 +127,7 @@ export interface CanDeployOptions {
   pactBrokerPassword?: string;
   output?: 'json' | 'table';
   verbose?: boolean;
+  to?: string;
   retryWhileUnknown?: number;
   retryInterval?: number;
   timeout?: number;
--- a/src/can-deploy.ts
+++ b/src/can-deploy.ts
@@ -2,6 +2,7 @@ import q = require('q');
 import logger from './logger';
 import spawn from './spawn';
 import pactStandalone from './pact-standalone';
+import { PACT_NODE_NO_VALUE } from './spawn';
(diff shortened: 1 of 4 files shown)


src/can-deploy.ts
```diff
@@ -124,11 +110,14 @@ export class CanDeploy {
 
 export default (options: CanDeployOptions) => new CanDeploy(options);
 
-export interface CanDeployOptions extends SpawnArguments {
-  participant?: string;
-  participantVersion?: string;
-  to?: string;
```
mefellows: Is there a reason why we removed this?

src/can-deploy.ts
```diff
@@ -124,11 +110,14 @@ export class CanDeploy {
...
-  latest?: boolean | string;
```
mefellows: ...and this?

src/can-deploy.ts
TimothyJones: These were moved into the `canDeployPacticipant` type.

I believe `--to=TAG` and `--latest=TAG` are synonyms, and are supported in the new interface by specifying a tag. 

I considered supporting the old API as well, but figured that it would be a lot of effort (and complexity) for little pay off (as you pointed out).

src/can-deploy.ts
TimothyJones: I've just realised you marked changes requested - did you want me to make it backwards-compatible?

mefellows: No, I think your other point answered this. We'll just leave them behind and upgrade Pact Node to the next major revision.

src/can-deploy.ts
mefellows: @bethesque is that correct about synonyms?

bethesque: Nope. --to=TAG is for the deployment environment. --latest=TAG is for specifying the version of the application that you're about to deploy.

eg 
```
can-i-deploy --pacticipant Foo --latest --to prod (can I deploy the latest version of Foo to prod)
can-i-deploy --pacticipant Foo --latest test --to prod (can I deploy the latest version of Foo with tag test to prod)
```

bethesque: https://github.com/pact-foundation/pact_broker-client#can-i-deploy

I would imagine most people would use the single pacticipant form of this command, but I have no way to know actual numbers.

kuangp: Hi,
Chiming in with my use-case regarding this change.
I initially attempted to use `canDeploy` but was disappointed that it only supported a single pacticipant so this would be a welcome change for me. I switched over to calling the cli `can-i-deploy` within a child process in my program in order to accomplish it currently. I am using the `latest` option in order to determine if some version of Foo is compatible with the latest marked prod version of Bar:  
`can-i-deploy --pacticipant Foo --version 123 --pacticipant Bar --latest prod`
but I also use the `to` option similar to the example given by @bethesque so I believe this option still needs to be distinctly supported

mefellows: Thanks @kuangp

bethesque: I'd say "determine compatibility with all the services in a deployment environment" rather than "configure compatibility".

TimothyJones: I can't find the documentation to update it (I can't see the source in the standalone repo, I can only see it on the releases page), but if they're not synonyms I think we should change the lines:

```
  -l, [--latest=[TAG]] 
      [--to=TAG] 
````

to

```
  -l, [--latest=[TAG]] 
  --to=TAG 
```

bethesque: The `--to TAG` is not required if you've specified the other pacticipants explicitly.

TimothyJones:
To avoid the changelog picking up commit messages which are no longer correct, I've reopened this as #190.

PR #178, opened 2019-08-19
refactor(spawn): Don't expose SpawnArguments, and split pact-util

TimothyJones:
Because of the index type in `SpawnArguments`, the options types are duplicated in pact-js. This is sometimes a [problem](https://github.com/pact-foundation/pact-js/issues/341) if the types aren't updated.

This PR ensures that the `SpawnArguments` implementation detail is not exposed, which removes the need to duplicate the types, and improves type safety (since the options types will no longer accept all dictionaries).

It also:

* Inverts `SpawnArguments` - it is now a union of all options types, instead of all options types inheriting it
* No longer exposes `SpawnArguments` outside pact-node
* Splits `pactUtil` into the three main concerns it dealt with - argument processing, binary spawning and removal, and knowing about the environment.
* Moves argument processing and binary spawning to be submodules of a `spawn` module.

Additionally, there is a small change to the logging format - where options are printed, they are now printed as JSON.


src/spawn/arguments.ts
```diff
@@ -0,0 +1,64 @@
...
+    args: SpawnArguments,
+    mappings: { [id: string]: string },
+  ): string[] {
+    return _.chain(args instanceof Array ? args : [args])
+      .map((x: SpawnArguments) => this.createArgumentsFromObject(x, mappings))
+      .flatten()
+      .value();
+  }
+
+  public stringifySpawnArguments(obj: SpawnArguments): string {
```
mefellows: Should we remove `stringifySpawnArguments` altogether? It doesn't appear to be used anywhere anymore.

src/spawn/arguments.ts
```diff
@@ -0,0 +1,64 @@
...
+  | MessageOptions
+  | PublisherOptions
+  | ServiceOptions
+  | VerifierOptions
+  | {};
+
+export const DEFAULT_ARG = 'DEFAULT';
+
+export class Arguments {
+  public fromSpawnArgs(
```
mefellows: Is there a case that `fromSpawnArgs` should actually be `toSpawnArgs` because it's job is to create
the spawn arguments?

src/spawn/arguments.ts
```diff
@@ -0,0 +1,64 @@
+const _ = require('underscore');
+const checkTypes = require('check-types');
+
+import { CanDeployOptions } from '../can-deploy';
+import { MessageOptions } from '../message';
+import { PublisherOptions } from '../publisher';
+import { ServiceOptions } from '../service';
+import { VerifierOptions } from '../verifier';
+
+export type SpawnArguments =
```
mefellows: What's the idea behind it being allowed to be an empty object?

TimothyJones: It made the tests significantly less noisy.

I thought it was safe to extend the type this way, because it's not exposed. But I haven't written much TypeScript, so it might be bad practice.

For now I'll put a comment in the definition.

src/spawn/arguments.ts
TimothyJones: Good catch! I changed it to `toArgumentsArray`, since it's taking `SpawnArguments` and turning them into `String[]`

mefellows:
🎉 awesome, thanks Tim! I'll confirm changes in Pact JS (probably Friday, possibly on a train ride) to remove the dupes. Nice!

PR #181, opened 2019-09-10
feat(canDeploy): resolve with output on success

### LINKED ISSUE #180: Resolve with output on canDeploy success
Would it be possible to resolve with the resulting output on a successful `canDeploy` call here: https://github.com/pact-foundation/pact-node/blob/master/src/can-deploy.ts#L111

This would allow the consumer to leverage the metadata returned by `can-i-deploy` that lists all known integrations and associated versions and also make it consistent with the error scenario which does return this information.

I would be happy to throw up a PR if this is approved

kuangp:
Closes #180


kuangp:
Hi @TimothyJones,

I wasn't sure what the best behavior is here. It makes sense that if the user specifies `--output=json` that we should resolve with the json object, however if the command fails then it rejects with an `Error` object whose string message is the json result which would be inconsistent with the successful output since the user would have to parse the error message in that case.

In order to maintain consistency in the output of `canDeploy` whether it succeeds or fails, we could reject with the output object|string directly instead of an `Error` object on failure. 

I've updated my change with the proposed behavior. Let me know what you think. Thanks!

TimothyJones:
The class is intended to be a wrapper for programatically calling the binaries, so I don't think that it makes sense for the user to specify the output type (especially as any parsing of the human-readable output might be brittle). Can you see a use case for someone programatically wanting the human readable output?

It's a common pattern to resolve with some data structure on success, and an `Error` on failure - but I see your point. I think it's worth distinguishing between failures because the check said no, and failures because something else went wrong with the binary. Maybe we can resolve with an object or an `Error`, depending on whether the check failed or something else went wrong.

kuangp:
@TimothyJones 
One reason that a user might want the human readable output is to directly display this output since it is pre-formatted. Apart from that I only attempted to follow the current documented behavior.  If it's deemed that the output should always be json when calling `canDeploy` programmatically then I can deprecate the `output` option and update the documentation appropriately if that is the direction we want to go?

As for distinguishing between an actual error and a failed deployment check, I'm not sure how to determine this? From my understanding, the binary exits with a non-zero code both when the deployment check fails and if the binary call fails unexpectedly unless there's a distinct exit code for a deployment check failure. (it also seems like the original `if` block on line 103 would crash attempting to parse and access properties on the output for the binary error scenario)

@bethesque do you have any thoughts or advice here?

bethesque:
I don't understand why there is any wrapper code at all to be honest. What is the use case?

bethesque:
Just found Tim's comment about programmatically calling the interface. If it's programmatic, then only JSON really makes sense, yes.

There is no distinct exit code for a "no you can't deploy" response. We could potentially add it if that would be helpful though.

TimothyJones:
@kuangp I thought I'd get this moving again by making the changes I suggested - let me know what you think. 

(Also, apologies for muddying the diff-  I made a mistake merging master in)

src/can-deploy.spec.ts
```diff
@@ -175,7 +175,10 @@ describe('CanDeploy Spec', () => {
       };
       const ding = canDeployFactory(opts);
```
mefellows: I know it's not related to this PR, but... why `ding`?

src/can-deploy.ts
```diff
@@ -87,7 +90,7 @@ export class CanDeploy {
     logger.info(
       `Asking broker at ${this.options.pactBroker} if it is possible to deploy`,
     );
-    const deferred = q.defer<string[]>();
+    const deferred = q.defer<any>();
```
mefellows: Should we declare a union type here - e.g. `string | CanIDeployResponse` type thing?

src/can-deploy.spec.ts
TimothyJones: I don't know. That's how it was when I got there.

src/can-deploy.ts
TimothyJones: That's a good point. Because we're just reading the json out of the broker cli, though, I'm not sure if we want to repeat the type (and keep it in sync with the ruby). I'm not very typescript savvy, so I'm not sure what the best practice would be.

src/can-deploy.ts
```diff
@@ -97,22 +100,28 @@ export class CanDeploy {
...
-        return deferred.resolve();
+      if (code === 0) {
+        logger.info(result);
+        return deferred.resolve(result);
       }
 
-      logger.error(`can-i-deploy did not return success message:\n${o}`);
-      return deferred.reject(new Error(o));
+      logger.error(`can-i-deploy did not return success message:\n${result}`);
+      return deferred.reject(result);
```
kuangp: Are we ok that this will not distinguish between deployment failure vs rejecting with an `Error` on binary call error?

kuangp:
Thanks @TimothyJones, your updates look good to me. Sorry, this got put on the back burner for me as I got busy with other work

src/can-deploy.ts
mefellows: I would declare an interface that matches the response type, so that it can be navigated via VS code etc. even in vanilla JS. Otherwise it's just an `any` object.

I think it's OK to repeat the type info here, because if it changes in Ruby it's a breaking, backwards-incompatible change (so it shouldn't be done without a major version bump).

A union type is the way to go I think, and I think it should actually be `Error | CanDeployResponse`

src/can-deploy.ts
mefellows: Good call, see above union type update.

src/can-deploy.ts
TimothyJones: This is actually a bit more complex, I think.

The current behaviour in this PR is one of:

* Output = json, Can deploy, resolve with the object returned from the broker
* Output = table, Can deploy, resolve with the human readable output
* Output = json, Can't deploy, reject with the object returned from the broker
* Output = table, Can't deploy, reject with the human readable output.
* An error occurred, reject with the error

This means that the type of the promise is `CanDeployResponse | string`, and a rejected promise might be `CanDeployResponse | string | Error` (but since the rejecting type isn't specified anywhere, we don't have the `Error` explicitly anywhere).

I don't particularly like that the promise might reject with something that isn't an `Error`, but also I don't think we should resolve the promise in the case where the server says you can't deploy.

Maybe the right thing to do is to make a custom error type for failed can-i-deploy checks?

mefellows: > I don't particularly like that the promise might reject with something that isn't an Error, but also I don't think we should resolve the promise in the case where the server says you can't deploy.

You make a good point. I like the custom `Error` types idea - what are you thinking?

mefellows: Did we want to pull this change in as is, get in https://github.com/pact-foundation/pact-node/pull/190 and then update to use the Error interface before releasing? Feels like the two PRs are overlapping, and want to avoid a mega PR change.

It's definitely looking _better_, though.

PR #187, opened 2019-10-04
chore(eslint): replacing tslint with eslint

flagoon:
Hey, it seems your tslint wasn't checking your code at all. When I plug in eslint with the same configuration, it found more that 200 errors. I tried to fix them, but I don't know your application good enough, so sometimes I had to use good, old eslint-disable rule. 

I mostly changes const something = require('something') to import something = require('something') as this way correct @types were imported. Also I installed bunch of @types for some libraries you are using in your project. Unfortunatelly not every library worked better with this and typescript, so I leave some things with eslint-disable. 

I added some modules' declarations to remove the error about @types, but it seems I could just leave const something...

I tried  to avoid changes in code as much as possible, but I had to comment out one parameter from object, because function using this object was using different interface. Probably you need to update this, but I don't know it as good to change it by myself (intstall.ts, function extract)

Tests are passing for me most of the time, but there is this test that if failing once in a while because of timeout. I believe it's implementation issue, not related with my changes.

Give me a sign if I need to correct something.
Thanks
Pawel


TimothyJones:
Oof! Sorry that it was more complex than expected, and thanks very much for seeing it through! Either I or @mefellows will get to this sometime in the next few days.

Thanks very much for the PR!

mefellows:
Oh dear. Thanks so much for this. I have a new 👶 so I'll probably be awake at some strange hour and will review. Much appreciated

TimothyJones:
This is excellent, thank you. My typescript fu is not strong so I might be missing something, but is there a reason that the following:

```
// eslint-disable-next-line @typescript-eslint/no-var-requires
const checkTypes = require('check-types');
```

can't be written as: 
```
import checkTypes = require('check-types');
``` 

?

flagoon:
Using const var = req('var') doesn't load the @types, which is bad, but in this case @types were showing the error. If you check the @types for this lib, there is a comment for this interface, that you can translate to: I don't know what I'm doing 😀. I decided to comment it out, and leave original code for safety.

flagoon:
Guys, can you check this error? As I said in PR, there is an issue with one test. It fails once in few tries because of timeout. Not only on my branch, but also on master branch. I don't think it's my fault.

mefellows:
Thanks for this @flagoon - there are a few minor things we'll have to work through over time (looking at you `eslint-disable`), but nothing that should prevent this coming in. We can improve the types over time.

PR #189, opened 2019-10-09
fix(verifier): Allow to use progress formatter during verification

### LINKED ISSUE #188: Expose formatting options for pact verification
## Software versions

-   **OS**: all 
-   **Library version**: 9.0.5
-   **Node Version**: all

## Description

Currently, it's not possible to use alternative formatting for provider verification. The underlying binary, [pact-provider-verifier](https://github.com/pact-foundation/pact-provider-verifier) allows alternative output from the verification process with the `format` option:

```
--format=FORMATTER     # RSpec formatter. Defaults to custom Pact formatter. Other options are json and RspecJunitFormatter (which outputs xml).
```

This option isn't exposed in pact-node, and should be. Relevant changes would be [here in pact-node](https://github.com/pact-foundation/pact-node/blob/master/src/verifier.ts) and [here in pact-js](https://github.com/pact-foundation/pact-js/blob/master/src/dsl/verifier.ts)

vgrigoruk:
Fixes #188

### DIFF
--- a/src/verifier.ts
+++ b/src/verifier.ts
@@ -141,9 +141,10 @@ export class Verifier {
 
     if (options.format) {
       checkTypes.assert.string(options.format);
-      checkTypes.assert.match(options.format, /^(xml|json)$/i);
-      options.format =
-        options.format.toLowerCase() === 'xml' ? 'RspecJunitFormatter' : 'json';
+      checkTypes.assert.match(options.format, /^(xml|json|progress)$/i);
+      if (options.format.toLowerCase() === 'xml') {
+        options.format = 'RspecJunitFormatter';
+      }
     }
 
     if (options.out) {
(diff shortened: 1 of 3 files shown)


mefellows:
Thanks - looks like the build failed at https://travis-ci.org/pact-foundation/pact-node/jobs/595583946#L121. Another reference to the rspec formatter in tests somewhere.

I'm ok with the change. It is also a breaking change, so would you mind please adding the text "BREAKING CHANGE" to the bottom of your commit message so that when we run our deployment process it automatically bumps the major version?

TimothyJones:
I pushed a commit fixing the build failure - unless I'm missing something, I don't think this is a breaking change, @mefellows?

src/verifier.ts
```diff
@@ -216,6 +217,6 @@ export interface VerifierOptions {
   timeout?: number;
   tags?: string[];
   monkeypatch?: string;
-  format?: 'json' | 'RspecJunitFormatter';
```
mefellows: @TimothyJones may bad. I thought this was breaking as it was replaced by `xml`. It looks OK now, so maybe I'm going crazy.

mefellows:
Merging - no breaking change.

PR #190, opened 2019-10-14
feat(can-i-deploy): allow multiple pacticipants to be specified to CanDeploy

TimothyJones:
BREAKING CHANGE: Options for CanDeploy have changed. Now, pacticipants are specified by an array of { name: <string>, latest?: <string | boolean>, version?: <string> }, allowing more than one pacticipant to be specified. You must specify one of latest
or version. If latest is `true`, the latest pact is used. If it is string, then the latest pact with that tag is used.

### DIFF
--- a/src/can-deploy.ts
+++ b/src/can-deploy.ts
@@ -2,6 +2,7 @@ import q = require('q');
 import logger from './logger';
 import spawn from './spawn';
 import pactStandalone from './pact-standalone';
+import { PACT_NODE_NO_VALUE } from './spawn';
 import * as _ from 'underscore';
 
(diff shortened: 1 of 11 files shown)


src/can-deploy.ts
```diff
@@ -11,29 +12,24 @@ export class CanDeploy {
...
+        ]),
+      ),
+    );
   }
 
   public readonly options: CanDeployOptions;
   private readonly __argMapping = {
-    participant: '--pacticipant',
-    participantVersion: '--version',
+    name: '--pacticipant',
```
mefellows: 👍

PR #193, opened 2019-10-22
fix(package.json): Move snyk to devDependencies

vgrigoruk:
- This PR fixes `package-lock.json`, which is  is currently out of sync with `package.json`
- Removes duplicated `snyk` entries in `package.json`
- Moves `snyk` to `devDependencies`, as it is not needed for runtime.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -79,6 +79,7 @@
       "version": "2.1.0",
       "resolved": "https://registry.npmjs.org/@snyk/cli-interface/-/cli-interface-2.1.0.tgz",
       "integrity": "sha512-b/magC8iNQP9QhSDeV9RQDSaY3sNy57k0UH1Y/sMOSvVLHLsA7dOi/HrPWTiLouyGqcuYzwjkz7bNbu8cwmVDQ==",
+      "dev": true,
       "requires": {
         "tslib": "^1.9.3"
       },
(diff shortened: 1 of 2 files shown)


TimothyJones:
I believe synk is meant to be a dependency, because it is used to patch vulnerable dependencies during npm install

mefellows:
That's what I've seen it doing. But I'm not sure that a) we should be enforcing that and b) it's caused 2 breakages in the past month alone. 

There was a chat on the Slack forum on this this morning (arguably should have been here). We can always bring it back, but right now it's causing an installation error (again) preventing the use of Pact JS (see https://github.com/pact-foundation/pact-js/issues/382).

TimothyJones:
Unless I’m missing something, this patch will also break installation, because synk protect will fail on install.

It definitely shouldn’t be a dev dependency. We could consider removing it entirely if it is causing problems.

vgrigoruk:
I'd suggest to remove snyk from package.json completely, if moving it to devDependencies doesn't solve the issue for other users.
Based on my experience, I'd prefer to avoid automatic patching of packages when `npm install` is executed on CI and during production deployment. I'd prefer to install what is defined in `package.json` + `package-lock.json`. I guess there are better ways to integrate snyk to the project.

mefellows:
So our `postinstall` script only installs the pact standalone tools.

`snyk protect` runs during our `prepare` phase (https://docs.npmjs.com/misc/scripts), which runs during local install (i.e. development). I think this is the behaviour we want.

> prepare: Run both BEFORE the package is packed and published, on local npm install without any arguments, and when installing git dependencies (See below). This is run AFTER prepublish, but BEFORE prepublishOnly.

Here is the `prepare` step running locally in the `pact-node` repo:

Screen Shot 2019-10-23 at 9 34 08 pm

The `prepare` step didn't run when installing the existing project in `/tmp`, just to be sure.

The [documentation](https://snyk.io/docs/cli-protect/) for the `protect` command states:

> Since running protect is the way to repeatedly apply patches, you should run it every time you reinstall your modules. Common integration points would be your CI/build system, your deployment system, and adding it as a post installation step in your package.json file (necessary if you consume this module via npm or yarn).

I'm with @vgrigoruk on this one - although I can see the upside, I'm not so sure I'm comfortable with enforcing snyk to run on target machines and would prefer to have it. I'm very much not comfortable with non-deterministic dependencies.

PR #196, opened 2019-10-28
Spawn binaries directly with arguments arrays, removing the need to quote options

### LINKED ISSUE #118: Single quote in consumer/provider names causes test failures
(Originally reported against `pact-js` [here](https://github.com/pact-foundation/pact-js/issues/160))

If the consumer or provider names contain single quote characters, the pact mock server doesn't spin up correctly, and connections are refused.

I've provided a branch that illustrates the problem in the Jest example for pact-js:

     git clone git@github.com:TimothyJones/pact-js.git
     cd pact-js
     git checkout single-quote-issue
     npm i && npm run dist && cd examples/jest/ && npm i
     npm run test

This happens because we escape the arguments [here](
https://github.com/pact-foundation/pact-node/blob/master/src/pact-util.ts#L27).

This escaping is necessary because we concatenate arguments to a single string [here](https://github.com/pact-foundation/pact-node/blob/master/src/pact-util.ts#L50).

If we instead passed arguments as an array,`child_process.spawn()` will correctly escape them (docs [here](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options)).  However, windows quoting isn't straightforward (which might be why we started doing the quoting in the first place).

It looks like the [cross-spawn](https://www.npmjs.com/package/cross-spawn) package would work as a replacement for `child_process` that is better for Windows.

I haven't provided a PR because I don't have a windows machine I can test on - but I'm happy to roll one up if someone else is able to test it.

TimothyJones:
This PR fixes #118 by spawning the binaries directly with `cross-spawn`, instead of going through a string passed to the shell. This should avoid issues to do with quoting on all platforms (including windows).

It also introduces a new concept, `CliVerbOptions`, which is internal to pact-node. This type is used where you have a verb (like `publish` in `pact-broker publish`) which must go before all other arguments.


TimothyJones:
Closing in favour of #202, which does not include the beta tag

PR #203, opened 2020-01-29
fix(spawn): none pactFileWriteMode throws an exception

mjlawrence83:
Whilst following the readme to figure out how to not commit to creating a pact file. I found that `none` was a valid option, but when setting it, an exception was being thrown from here.

I will patch it locally for now as I've proven this change works. However, I currently don't know if it will break your tests?

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -58,7 +58,7 @@ export class Server extends AbstractService {
     }
 
     checkTypes.assert.includes(
-      ['overwrite', 'update', 'merge'],
+      ['overwrite', 'update', 'merge', 'none'],
       options.pactFileWriteMode,
     );
 


mefellows:
@mjlawrence83 may I ask to understand the use case a little better? Why would you not want to commit to creating the pact file?

@bethesque `'none'` appears to be an undocumented but valid [option](https://github.com/pact-foundation/pact-mock_service/blob/a0e24a23c1048d50069d38ac0b0c85242344209d/lib/pact/stub_service/cli.rb#L51). Is it a valid thing to want to do?

bethesque:
The reason it's used is that the same code is used for the mock service and the stub service. The 'none' tells the mock service not to write a pact file when it's running in stub mode. It's not an option that's meant to be set by the end user.

TimothyJones:
I agree that I'd like to hear more about the use case - if there's a good reason for needing this option, then we could accept this PR, and make `none` part of the documentation for the CLI.

mefellows:
The stub service does require a pact file which isn't always that helpful,
so perhaps one use case is to be able to extend a set of pact tests
programmatically for eg end to end UI tests

Your could add more data/boundary tests that you may not want in the
serialised contract.

But that's hypothetical, I'd generally use other tools for that job

mjlawrence83:
Hi all, I've introduced Pact across our existing Jest unit testing suite...I'm also currently integrating it into our Cypress driven testing.

I have not gone down the route of creating a dedicated set of Pact tests which just exercise the contract & their types. Instead, For the consumers, I've replaced all our original mocks and opted to use Pact as a mock service to provide responses for the unit tests. This means that I'm providing pact with the exact responses (with correct data) and I'm continuing & asserting the tests using the data Pact returns. I'm seeing this a way of trying to achieve integration testing without the actual services. 
For the providers, providing we setup the test database with the same expected data declared in the consumers tests, all should pass.

The reason I require the `none` is because the pact providers are set up at the start of every Jest test suite run, and at the moment always `update` the existing Pact files. This is proving to be a pain for some developers which only runs a subset of the test suite. After running a subset, it sometimes changes the Pact files causing their git status to flag up pact file changes which then confuses them.
A cheap workaround I've experimented with here is to only create generate the Pact files when `npm test` aka the full suite, is ran to reduce confusion. I set a environment variable to switch the pactFileMode to either `none` or `update`.

Another problem I'm experiencing beyond the scope of this PR, is if code uses promise.all([]), I can't guarantee the order of the returns, which causes the Pact file to change as the ordering of the commands can be different between runs. Maybe this scenario is connected to why I see the pact file change?

This explanation may also cover what I've raised in https://github.com/pact-foundation/pact-js/issues/402

bethesque:
> This is proving to be a pain for some developers which only runs a subset of the test suite.

This is what the `merge` option was created to support. I had a really long test suite, and when I only ran one spec, it deleted all the interactions except for the one that I had run. I found this annoying, so I created an option to just merge in any executed interactions with the existing file, based on the provider state and description.

If you use the `merge` option though, you have to make sure that at the start of a full test run, you delete all the pact files yourself though.

Most people don't commit their pact files however - they publish them to the Pact Broker (or some other repository) so it's not usually an issue.

TimothyJones:
Thanks for your work on this, it provoked some interesting discussion.

We're going to politely decline this PR - we think it's more likely to lead to a misuse or misunderstanding of the tooling. I'll make sure we update the relevant parts of the documentation (eg `merge`) so that we avoid misleading users in the future.

Apologies for not merging your PR - we'd generally like to accept community contributions. I hope this doesn't discourage you from making contributions in the future.

Please feel free to comment if you have any further questions, or if you'd like me to elaborate. Thanks again, and apologies for not accepting this one.

PR #206, opened 2020-02-18
feat: update standalone to 1.80.0

bethesque:


bethesque:
Just automated the PR process. WINNING AT LIFE.

bethesque:
Not sure why the checksum check is failing. It may be because the standalone release failed midway due to a network error, and I re-ran it. I'm doing a patch release, and see if that works.

PR #209, opened 2020-02-28
Fix/decompress vulnerability

mefellows:
Fixes critical vulnerability in `decompress`.

The library I used is well maintained but doesn't support stripped folders from the output so I've adjusted to deal with that.

I also fixed some linting whilst in there so I haven't squashed those commits.


uglyog:
LGTM, I assume it still needs to be tested?

mefellows:
It's been tested - appveyor/travis are happy. I'll merge in now.

PR #212, opened 2020-03-06
fix: set strictSSL on custom downloads from NPM config. Fixes #211

### LINKED ISSUE #211: Postinstall seems to ignore strict-ssl=false
We are using pact behind a corporate firewall. In view of https://github.com/pact-foundation/pact-node/issues/88 we have the following pact-related entries in our package.json:

```
  "config": {
    "pact_binary_location": "https://artifactory.ourcompany.com.au/artifactory/pact-github-repo/download/v1.81.0/"
  },
...
  "devDependencies": {
    "@pact-foundation/pact": "^9.6.0",
...
```
Our .npmrc contains
```
strict-ssl=false
cafile=/usr/local/ourcompany/certificates/ourcompany-proxy-internal.cer
```
which works for all packages - except for pact-foundation. We get:
```
Installing Pact Standalone Binary for darwin.
Downloading Pact Standalone Binary v1.73.0 for platform darwin from https://artifactory.ourcompany.com.au/artifactory/pact-github-repo/download/v1.81.0/pact-1.73.0-osx.tar.gz
Error: Error while installing binary: Postinstalled Failed Unexpectedly: Error downloading binary from https://artifactory.ourcompany.com.au/artifactory/pact-github-repo/download/v1.81.0/pact-1.73.0-osx.tar.gz: Error: Error while installing binary: Error: self signed certificate in certificate chain
```
We can curl https://artifactory.ourcompany.com.au/artifactory/pact-github-repo/download/v1.81.0/pact-1.73.0-osx.tar.gz, so we know the URL is correct. 

It appears that the postinstall script https://github.com/pact-foundation/pact-node/blob/master/standalone/install.ts does not use the strict-ssl=false setting.

mefellows:
Uses `libnpmconfig` to cascade through the various npm config layers and set the property on the download object.

Didn't like exporting the function in order to test it, but I needed a few tests around it at the time. Perhaps they can go now?

### DIFF
--- a/standalone/install.spec.ts
+++ b/standalone/install.spec.ts
@@ -4,7 +4,7 @@
 import * as fs from 'fs';
 import * as path from 'path';
 import * as chai from 'chai';
-import { BinaryEntry, Config, useStrictSSL } from './install';
+import { BinaryEntry, Config } from './install';
 
 const expect = chai.expect;
 
@@ -18,27 +18,6 @@ describe('Install', () => {
       delete require.cache[key];
     }
   });
-
-  describe('#useStrictSSL', () => {
-    afterEach(() => {
-      delete process.env['npm_config_strict-ssl']
-    })
-    describe('when "strict-ssl" is specified in an NPM configuration', () => {
-      it('returns a true', () => {
-        process.env['npm_config_strict-ssl'] = "true"
-        expect(useStrictSSL()).to.be.true
-      })
-      it('returns a false', () => {
-        process.env['npm_config_strict-ssl'] = "false"
-        expect(useStrictSSL()).to.be.false
-      })
-    })
-    describe('when "strict-ssl" is not specified in any NPM configuration', () => {
-      it('returns true', () => {
-        expect(useStrictSSL()).to.be.true
-      })
-    })
-  })
 
   function createConfig(): Config {
     return require('./install').createConfig();
--- a/standalone/install.spec.ts
+++ b/standalone/install.spec.ts
@@ -4,7 +4,7 @@
 import * as fs from 'fs';
 import * as path from 'path';
 import * as chai from 'chai';
-import { BinaryEntry, Config } from './install';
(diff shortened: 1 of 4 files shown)


standalone/install.spec.ts
```diff
@@ -4,7 +4,7 @@
 import * as fs from 'fs';
 import * as path from 'path';
 import * as chai from 'chai';
-import { BinaryEntry, Config } from './install';
+import { BinaryEntry, Config, useStrictSSL } from './install';
```
TimothyJones: `useStrictSSL` doesn't exist any more in this PR, I think? It's certainly not used in this file. 

I tend to think that exporting things only for testing implies that the module cut points aren't right yet. Last time I was looking at modifying `install.ts`, I was thinking about refactoring it out to separate the different concerns. I didn't, because it was going to be substantial work that would be hard to test.

Pragmatically, I liked this change more when it had the tests, and I would like it even more if the config parts of `install.ts` were extracted to their own module. What do you think?

mefellows: Oh crap, yes I did remove that in the end. I'm surprised something didn't get upset about it (or maybe it did).

> I tend to think that exporting things only for testing implies that the module cut points aren't right yet.

Agree (and hence my initial concern), but for a minor fix didn't want to refactor either! But also wanted tests.

To be honest, and this is definitely outside the scope of this, but the whole 

1. Have to `npm t` before `npm i` is a bit strange
1. Have to `npm t` before `npm run release` (which also runs `t` anyway)

is a bit strange, and counter intuitive. I should be able to pull repo, `npm it` and be set 👌. I got tripped by this on the last release with the "checksum invalid" issue, only to realise that I needed to re-run all of the tests (which bring down all of the OS specific binaries).

I don't want all of this of course to get in the way of fixing a bug, so perhaps we create an issue, mark it as "help wanted" / "good first issue"?

mefellows: (I've fixed that broken import by the way)

PR #215, opened 2020-03-15
feat: update standalone to 1.82.0 for the cors allow auth headers feature

bethesque:


bethesque:
See https://github.com/pact-foundation/pact-mock_service/pull/121

PR #217, opened 2020-03-25
feat: add consumer version selectors and pending pacts to verification

mefellows:
Implements consumer version selectors (see https://github.com/pact-foundation/pact_broker/issues/307) and pending pacts (see ).

The biggest complication here is that the selectors are specified as JSON strings on the CLI - not something we're well prepared for.

In terms of design options I had a few, most notably:

1. Modify `verifier.ts` to do the JSON marshaling, but that then meant I needed to do nasty things to hide the true option that is passed to the `Arguments` class.
2. Enhance the `SpawnArguments` type to signal that certain properties were basic "string" values vs "JSON" values - but decided against that large refactor.
3. Detect an argument value - if it is an object, marshal to JSON, otherwise fall back to existing behaviour. This is the path I chose, because it was so simple to implement.

I also updated the types in `Arguments` because the values were set as strings, but that is actually an illusion - in reality they are an interface type (`SpawnArguments`), that could have values as primitives (strings, numbers etc.) or in this new case - an interface. 

This was a fairly simple change, but I was concerned about making that class any more complicated, so we may want to revert it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,7 @@ pact.verifyPacts({
 | `providerBaseUrl`           | true      | string  | Running API provider host endpoint.                                                                        |
 | `pactBrokerUrl`         | false     | string  | Base URL of the Pact Broker from which to retrieve the pacts. Required if `pactUrls` not given.                                                                        |
 | `provider`                  | false     | string  | Name of the provider if fetching from a Broker                                                             |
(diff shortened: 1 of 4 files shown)


bethesque:
Also docs.pact.io/selectors

mefellows:
Thanks Beth - updated the links.

TimothyJones:
Nice work, @mefellows !

PR #219, opened 2020-04-06
feat: deprecate consumerVersionTag and providerVersionTag. Fixes #218

### LINKED ISSUE #218: Please rename consumerVersionTag to consumerVersionTags
`consumerVersionTag` gives the impression you should only be verifying one (even though it takes an array)

`providerVersionTags` the same (though the main usecase would just be one)

mefellows:
* Adds `consumerVersionTags` and `providerVersionTags` as options that _only_ accept arrays. I think accepting a string or array is unnecessary.
* Logs deprecation warning if `consumerVersionTag` and `providerVersionTag` is used
* Supports both options until a major version bump.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,8 +269,8 @@ pact.verifyPacts({
 | `pactBrokerUrl`         | false     | string  | Base URL of the Pact Broker from which to retrieve the pacts. Required if `pactUrls` not given.                                                                        |
 | `provider`                  | false     | string  | Name of the provider if fetching from a Broker                                                             |
 | `consumerVersionSelectors`        | false     | ConsumerVersionSelector\|array  | Use [Selectors](https://docs.pact.io/selectors) to is a way we specify which pacticipants and versions we want to use when configuring verifications.                                                         |
-| `consumerVersionTag`        | false     | string\|array  | Retrieve the latest pacts with given tag(s)                                                        |
-| `providerVersionTag`        | false     | string\|array  |  Tag(s) to apply to the provider application |
+| `consumerVersionTags`        | false     | string\|array  | Retrieve the latest pacts with given tag(s)                                                        |
+| `providerVersionTags`        | false     | string\|array  |  Tag(s) to apply to the provider application |
 | `pactUrls`                  | false     | array   | Array of local pact file paths or HTTP-based URLs. Required if _not_ using a Pact Broker.                  |
 | `providerStatesSetupUrl`    | false     | string  | URL to send PUT requests to setup a given provider state                                                   |
 | `pactBrokerUsername`        | false     | string  | Username for Pact Broker basic authentication                                                              |
--- a/src/verifier.spec.ts
+++ b/src/verifier.spec.ts
@@ -1,7 +1,10 @@
 import path = require('path');
 import chai = require('chai');
 import chaiAsPromised = require('chai-as-promised');
-import verifierFactory, { VerifierOptions } from './verifier';
+import verifierFactory, {
+	VerifierOptions,
+	DeprecatedVerifierOptions,
+} from './verifier';
 
 const expect = chai.expect;
 chai.use(chaiAsPromised);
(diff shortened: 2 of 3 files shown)


TimothyJones:
Unless I'm missing something, I can't see where the two options are used by the lower layer.

It would be good if we didn't muddy the type definition with the two types too - maybe we should have a separate `DeprecatedVerifierOptions` which is unioned in? That way we could send only the new complete type down to lower layers after validation.

>  I think accepting a string or array is unnecessary.

Shouldn't we be permissive in what we accept, and restrictive in what we send? ;)

mefellows:
> Unless I'm missing something, I can't see where the two options are used by the lower layer.

You are right, I didn't push that line up inexplicably. Fixed.

> It would be good if we didn't muddy the type definition with the two types too - maybe we should have a separate DeprecatedVerifierOptions which is unioned in? That way we could send only the new complete type down to lower layers after validation.

Would you mind showing how that would help? Unioning it in will still show it up in the type interface AFAIK. I thought about using decorators but that may not be portable, and won't help non-TS/JS users anyway. So the `warn` was the best I could come up with. Agree, it's a bit shit - alternatively I bump a major version and just remove completely.

>    I think accepting a string or array is unnecessary.
> Shouldn't we be permissive in what we accept, and restrictive in what we send? ;)

With that argument, we should accept strings for everything and do the coercion behind the scenes. But seriously, do you think we should do both? It's a pluralised property now, so for me, it makes sense for it just to be an array. But if you feel strongly...

TimothyJones:
> Would you mind showing how that would help? 

You're right that the compiler would see the unioned type, but for users reading the type, I think it's clearer to be able to see the deprecated options separately.

I think it's largely readability / layer separation. I think there are two types:

* What we expose to callers (kind of the `DTO`)
* What we pass to the lower layer (kind of the domain entity)

What's convenient to keep each layer simple might be different. For example, it's not convenient if the layer that consumes the options knows that there once was another way of specifying the options. Similarly, it's not convenient if the lower layer knows that you also can provide a string instead of an array (I've been unrelatedly thinking about clearing this up - kind of marshalling a user-friendly structure into a clear easily mappable domain config structure).

> With that argument, we should accept strings for everything and do the coercion behind the scenes. 

Well, not exactly :) I think APIs should only accept not-quite-as-written if it is completely unambiguous. But also- yes. I do think that we should coerce where it is clear what the user meant.

> It's a pluralised property now, so for me, it makes sense for it just to be an array.

If you're saying it's plural so we shouldn't accept a single string, I think the same argument could be made for not accepting `consumerVersionTags: ["oneTag"],` (which I don't think is a strong argument).

My view is that APIs should be hard to misuse, and that programmers shouldn't have to do work because the computer exists. In this case the "extra work" is putting in `[]`, so I would say my summary is:

> But seriously, do you think we should do both?

I do.

> But if you feel strongly...

I don't :)

TimothyJones:
All the discussion above is why I didn't get to this change yet, though :)

Btw, I can see that you pushed again, but for some reason I can't see the change. Maybe it's github caching the diff or something.

mefellows:
OK, given the change was so simple I've gone with your suggestions Tim. I'm not 100% liking the `VerifierOptions & DeprecatedVerifierOptions` littered everywhere, but I think having a reminder will mean we actually will get around to deprecating it.

I could have gone with something like `export type CompleteVerifierOptions = VerifierOptions & DeprecatedVerifierOptions` but I wasn't sure that was going to have the intended interface result. 

Feel free to change as you will if you're picking up the WIP pacts anyway.

TimothyJones:
Ah - making it the way I was thinking of was messier than I was thinking of - I can see why you were asking for an example 😂 

I was hoping we could export a type which was a union of deprecated and current (which I've now done), and inside the constructor convert it to just the current options.

I'd like to rethink this whole layer at some point. It seems a lot heavier than it needs to be, but I've left that for a rainy day.

PR #222, opened 2020-04-07
fix: correcting the cafile setup for postinstall download

oxlade39:
fix for issue discovered whilst trying to use pactjs

https://github.com/pact-foundation/pact-js/issues/427

CA config does not work when a cafile is specified in .npmrc


mefellows:
Thanks for this! I'll tidy up and get a release out today.

PR #227, opened 2020-06-03
fix: update standalone to 1.85.0

bethesque:
Might fix https://github.com/pact-foundation/pact-js/issues/446


mefellows:
Thanks Beth.

PR #238, opened 2020-08-26
Migrate Logger to Pino

neezer:
Extracted and refined based on feedback from https://github.com/pact-foundation/pact-js/pull/489


src/server.ts
```diff
@@ -37,6 +38,7 @@ export class Server extends AbstractService {
 
 		if (options.log) {
 			options.log = path.resolve(options.log);
+			setLogPath(options.log);
```
neezer, line 41: @TimothyJones [wrote](https://github.com/pact-foundation/pact-js/pull/489#issuecomment-680647257)

> Ideally, we'd set them to append to the same place, but my gut is that it will produce flakey behaviour on windows, which doesn't like the same files being held open by multiple processes.

This change would write the logs to the same file. I don't have a Windows machine handy to test on, but I figured this might be a good opportunity to test your theory and see if it's actually an issue, [considering](https://github.com/pact-foundation/pact-js/pull/489#issuecomment-679937389)

> ... as of last Friday, I now have a windows dev machine on loan.

If it doesn't work, I can amend the PR to remove this change and we can brainstorm on some other way of configuring this behavior.

TimothyJones: Thanks for this!

We have windows builds in appveyor- they've been flakey in the past because the environment they provide sometimes changes, but they should (as of just now) be green again in master.

However, it looks like the failure on this PR is real, while clearing out the logging directory during tests. My guess is that a handle to the file gets left open by pino? You can see the logs linked in the appveyor checks at the bottom of this PR.

Happy for you to take a look, otherwise I'll have some time to check it out on my windows machine on Sunday.

TimothyJones: Oh my goodness, it has been a month since I last commented here. Apologies for taking so long!

Unfortunately, I wasn't able to try this out in a windows environment (unrelated issues because of not having admin access on the machine I was borrowing prevented development).

This means I think we'll need to use the appveyor env to ensure this change works - the current version of this PR isn't passing in appveyor, but I'm not sure if that's because of appveyor config or the change. Master does pass, though.

Perhaps we should find a time to look in to it together?

neezer: @TimothyJones Sure, be happy to try and jam on this together. How do you want to coordinate scheduling?

TimothyJones:
@neezer  I managed to get the current test to pass. I'm not sure if that's enough to revert the debug commit, but I ran out of time to look further this evening.

Also sorry about the force push - I was reverting accidentally included everything from the current head of master when I was checking it out locally.

Feel free to hit me up on the Pact Foundation slack (pact-foundation.slack.com). Let's see if we can get this in over the next week or two

neezer:
@TimothyJones wrote

> Feel free to hit me up on the Pact Foundation slack (pact-foundation.slack.com). Let's see if we can get this in over the next week or two

Appears I need an invitation to join; mind sending me one? This random email address should reach me: photesthesis_geospiza@aleeas.com

TimothyJones:
@neezer Apologies for the many many months it took to get this in, but I've finally had the time to get to it. Thanks again for originally looking into this and kicking off the process! I'll make sure we're faster in future!

mefellows:
Thanks for getting it in also Tim!

neezer:
@TimothyJones All good! Glad it finally got in. I've been busy too, so apologies right back at ya for not championing more consistently. 🤷‍♂️

PR #241, opened 2020-09-25
fix(install) Resolve yarn cafile from the project root

eyezak:
When using a .yarnrc file and a local cacert, the ca file is assumed to be an absolute path
This change resolves the path relative to process.env.INIT_CWD when present


TimothyJones:
Thanks for the PR! 

My understanding is that `INIT_CWD` is the location that the user invoked the `npm` (or yarn) script from. This might not be the root of the project, so I think this change might result in unexpected behaviour for the user.

Could you give a bit more context for this change? Would `path.resolve(cp)` solve the issue more completely?

PR #242, opened 2020-10-05
fix: Correct the config for using custom ca file

grahamhar:
When behind a proxy that implements SSL interception the CA file was not being set correctly.

I couldn't think of a way to add tests for this but if a way can be suggested I'm happy to implement them. I have however tested this locally behind an intercepting proxy.

This could possibly fix https://github.com/pact-foundation/pact-js/issues/485


grahamhar:
I've tried re-running the travis tests and a different set failed this time, I believe these failures are not related to my changes

mefellows:
You're right, they look to be failing intermittently - thanks for fixing this. No worries about the test, we can take it from here.

Issue #236, opened 2020-08-26
Provide mechanism to skip binary download/install during "npm install"


uglyog:
For context, see https://github.com/cypress-io/cypress/issues/1005

mefellows:
Thanks Ron, should be an easy thing to do. Will look at it later this week.

TimothyJones:
Fixed in v10.11.0

PR #243, opened 2020-10-19
feat(binary): Allow skip binary downloading

bangn:


standalone/install.spec.ts
```diff
@@ -92,4 +92,20 @@ describe('Install', () => {
...
+
+		beforeEach(() => {
+			process.env.PACT_SKIP_BINARY_INSTALL = 'true';
+		});
+
+		afterEach(() => {
+			process.env = { ...OLD_ENV };
+		});
+
+		it.only('Should not download it', () => {
```
TimothyJones, line 107: I think this might be better as `it`, not `it.only`.

TimothyJones:
Thanks for this!

Could you tell us a bit about the use case? Does the [download location setting](https://github.com/pact-foundation/pact-node#pact-download-location) cover it?

standalone/install.spec.ts
bangn: My bad :face_palm:. Fixed 3efffc5c4bb08fe

bangn:
> Thanks for this!
> 
> Could you tell us a bit about the use case? Does the [download location setting](https://github.com/pact-foundation/pact-node#pact-download-location) cover it?

Sorry, I should mention this is to fix issue #236

bangn:
In addition, [download local setting](https://github.com/pact-foundation/pact-node#pact-download-location) only change the location of the binary.
This feature is to ignore the binary totally. The usecase I can think of (please
correct me if im wrong) is that the binary is already downloaded so user would
like not to download it again.

mefellows:
Long time no see @bangn!

TimothyJones:
Cool. That's a good rationale. We should add some documentation to the readme too before we release. Would you be able to do that in this PR?

TimothyJones:
Also, sorry to nitpick - commits starting with `feat` or `fix` end up in the release notes - could you do a force push that changes the `feat` in `feat(binary): Remove only from the test` to something like `chore`?

bangn:
> Long time no see @bangn!

Hi @mefellows, long time no see :D

@TimothyJones

> could you do a force push that changes the feat in feat(binary): Remove only from the test to something like chore?

Done!

> We should add some documentation to the readme too before we release. Would you be able to do that in this PR?

I'll do that. Thank you

TimothyJones:
Releasing this in 10.11.0

PR #265, opened 2021-02-24
fix: 🐛 only download binary if PACT_SKIP_BINARY_INSTALL is true

danymarques:
Only download Pact binaries if the environment variable PACT_SKIP_BINARY_INSTALL has the value 'true'.

Actually, we only check if the environment variable exists. This means we don't care about its value.


mefellows:
Thanks Dany!

I'm confused a bit, though. The whole point of that env var is to avoid downloading the binary. Can you please provide more context?

Also, there are a lot of other changes in the PR that are seemingly unrelated, could you please comment on those?

TimothyJones:
@mefellows : Sorry, I should have filled you in. @danymarques and I have been working together on this one, which is why he didn't provide the context in the PR.

It's a bit of a yak shave:

1) The key feature here is that the old code didn't match the documentation (Which says it will skip if `PACT_SKIP_BINARY_INSTALL=true`, but actually it would skip if any value were set - eg `PACT_SKIP_BINARY_INSTALL=false` would still skip)
2) While testing this simple change, Dany noticed that the test would never fail, which we fixed
3) He also noticed that it was checking the wrong thing - whether the promise resolved, not whether the binary was skipped. So, this PR includes a rewrite of the tests for install.ts that better check all the things.
4) To avoid some issues with mocha and mocking the file system (I think, I don't remember the details), the test has been migrated to Jest (if we finish this migration, this will help with the log levels leaking between tests - something that I can't see a way to fix in mocha).
5) There is also some clean up of the mocha settings, moving them out of the script in package json and into config

To be clear, it's all Dany's work, I'm not responsible for anything other than being across it :+1: And neglecting to tell you 👎 

:) 🚀

mefellows:
Haha ok - yes, I saw Jest and config files and all these things. That helps - I'll disappear.

@danymarques thank you very much!

danymarques:
We forget to give some context, that's true :-) Sorry @mefellows 
Thanks for the collaboration one more time @TimothyJones

mefellows:
Don't be sorry @danymarques, I'm super grateful to you both 🙏  .

Issue #264, opened 2021-02-11
Support installation on Mac M1 ARM (using rosetta2)


pedropaf:
I'm trying to install PACT for testing, I get the following error:

```
arch -x86_64 npm install @pact-foundation/pact-node
npm ERR! code EBADPLATFORM
npm ERR! notsup Unsupported platform for @pact-foundation/pact-node@10.11.8: wanted {"os":"darwin,linux,win32","arch":"x64,ia32"} (current: {"os":"darwin","arch":"arm64"})
npm ERR! notsup Valid OS:    darwin,linux,win32
npm ERR! notsup Valid Arch:  x64,ia32
npm ERR! notsup Actual OS:   darwin
npm ERR! notsup Actual Arch: arm64
```

I have tried with node 12 and node 15. / I have tried to specify `arch x86_64` to run on rosseta2, but I think the issue is the call to the binary form js, would need to specify that flag if the architecture is darwin / arm64 pass the arch flag when calling the binary file to run it through rosetta2.

Any other workarounds that you suggest?

Thanks,
Pedro

TimothyJones:
So, we don’t currently have M1 binaries (or an M1 machine to test on, but I’m sure that will change). Travelling ruby has just been updated, so we might be able to make M1 binaries with that.

I think this error is coming from npm, not from pact-node (although it is following the restriction in pact-node’s package.json). From the npm docs, it looks like you’re running the arm version of npm. I think you’ll need to run the x86 version of node to install it- sounds like this is possible/easiest through running your terminal through Rosetta (although I may be wrong, I have only just started reading about it).

TimothyJones:
It's very possible that there will be issues with the way we are invoking the binary at the moment, but I am pretty sure the error you're currently getting is from npm deciding whether or not to let you install the package, based on comparing the value of `process.arch` to pact-node's `cpu` field in package.json.

If you have the time, it might be easiest to jump on slack together and see what we can do to get this going- I'm around tomorrow morning AEDT (UTC+11) -and plenty of other times if that doesn't work.

pedropaf:
Thanks, @TimothyJones for the update, I'm happy to jump into a call on GMT timezone.

fimak:
@pedropaf @TimothyJones 
How it was, did you find a solution?

obiSerra:
hello, there is any news on this issue? I'm having the same problem

obiSerra:
[...]

PR #271, opened 2021-03-15
fix: Standalone download and installation for Apple m1

obiSerra:
The purpose of this PR is to enable the download and installation of the Standalone binaries for Apple m1 chips (os: `darwin`, arch: `arm64`).

Binaries for Mac OS seem to be already compatible and only a check made via the package.json prevents the download and installation.

I also added a step for downloading and checking the checksum, as the process was failing because the checksum file was not downloaded automatically


konalegi:
@obiSerra do you know, is this message somehow could be fixed by your PR? 
I have the following issue with docker on m1
```
#9 26.86 error @pact-foundation/pact-node@10.11.11: The CPU architecture "arm64" is incompatible with this module.
```

TimothyJones:
@konalegi The error message is correct, because we don't have binaries for M1 yet. See #264 for details and a workaround.

I'm going to close this for now, because we can't merge it before we support M1.

konalegi:
@TimothyJones Those workarounds don't work for me, because I'm running inside docker which is under arm64. I guess I have to wait for m1 support.

TimothyJones:
@konalegi: We changed direction on this, and pact-node@10.13.1 / pact-core@11.1.0 have this workaround included.

PR #299, opened 2021-06-23
WIP: Use ffi bindings for the verifier

TimothyJones:
This replaces the ruby verifier core with one that uses the the FFI bindings.

It's a breaking change, and I haven't documented it for the changelog (or the documentation) so we can't merge it yet. However, it does add a straightforward pattern that we can potentially use for the other ffi bindings.


src/verifier/types.ts
```diff
@@ -0,0 +1,33 @@
...
+  consumerVersionTags?: string | string[];
+  providerVersionTags?: string | string[];
+  providerStatesSetupUrl?: string;
+  publishVerificationResult?: boolean;
+  providerVersion?: string;
+  enablePending?: boolean;
+  // Todo in FFI
+  includeWipPactsSince?: string;
+  consumerVersionSelectors?: ConsumerVersionSelector[];
+  logLevel?: LogLevel;
```
mefellows, line 28: So a more general point is that we have access to these logging facilities now: https://github.com/pact-foundation/pact-reference/blob/master/rust/pact_mock_server_ffi/src/lib.rs#L88-L97.

There is still an outstanding PR to enable logging into it from a remote interface (this was the [WIP PR](https://github.com/pact-foundation/pact-reference/pull/105) to demonstrate the concept)

This property (or something like it) should be supported. We have a few choices, I think we were trending towards one of the following:

1. Log everything through the rust interface and print to std out
2. Log everything through the rust interface and log to file
3. Log everything through the rust interface (via `log_to_buffer`, `log`) but fetch the buffer at the end of a run per mock server via `fetch_log_buffer` and control the output via pact-core

I think 3 would be ideal.

test/verifier.integration.spec.ts
```diff
@@ -15,9 +16,12 @@ describe('Verifier Integration Spec', () => {
   const providerStatesSetupUrl = `${providerBaseUrl}/provider-state`;
   const pactBrokerBaseUrl = `http://localhost:${PORT}`;
   const monkeypatchFile: string = path.resolve(__dirname, 'monkeypatch.rb');
```
mefellows, line 18: Is this test here to show the logger output? Or should it go?

src/verifier/nativeVerifier.ts
```diff
@@ -0,0 +1,62 @@
...
+const VERIFICATION_FAILED = 1;
+// 2 - null string passed
+// 3 - method panicked
+const INVALID_ARGUMENTS = 4;
+
+export const verify = (opts: VerifierOptions): Promise<string> => {
+  const verifierLib = getVerifierLib();
+  // Todo: probably separate out the sections of this logic into separate promises
+  return new Promise<string>((resolve, reject) => {
+    // Todo: Does this need to be a specific log level?
```
mefellows, line 17: This accepts the name of an environment variable that can be set.

e.g. if you did `LOG_LEVEL=trace npm run ...` it would print rust level trace. Or put another way, the environment variable is what actually sets the log level.

src/verifier/nativeVerifier.ts
```diff
@@ -0,0 +1,62 @@
...
+            break;
+          case INVALID_ARGUMENTS:
+            logger.pactCrash(
+              'The underlying pact core was invoked incorrectly.'
+            );
+            reject(new Error('Verification was unable to run'));
+            break;
+          default:
+            logger.pactCrash(
+              'The underlying pact core crashed in an unexpected way.'
```
mefellows, line 54: Should we consider a terraform-like output here? If this happens, it's definitely a bug that needs to be reported, and we should make it as easy as possible to report.


```
!!!!!!!!!!!!!!!!!!!!!!!!!!! TERRAFORM CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!

Terraform crashed! This is always indicative of a bug within Terraform.
A crash log has been placed at "crash.log" relative to your current
working directory. It would be immensely helpful if you could please
report the crash with Terraform[1] so that we can fix this.

When reporting bugs, please include your terraform version. That
information is available on the first line of crash.log. You can also
get it by running 'terraform --version' on the command line.

[1]: https://github.com/hashicorp/terraform/issues

!!!!!!!!!!!!!!!!!!!!!!!!!!! TERRAFORM CRASH !!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

src/logger.ts
```diff
@@ -3,21 +3,48 @@ import pino = require('pino');
...
   });
+  pinoLogger.pactCrash = (message: string) => {
+    pinoLogger.error(pactCrashMessage(message));
+  };
+  return pinoLogger;
+};
+
+const pactCrashMessage = (
+  extraMessage: string
+) => `!!!!!!!!! PACT CRASHED !!!!!!!!!
```
mefellows, line 26: 👨‍🍳  😗  (this is the closes to chef-kiss GH has...)

src/verifier/nativeVerifier.ts
mefellows: Nevermind - see above!

src/ffi/internals/index.ts
```diff
@@ -0,0 +1,57 @@
...
+import ffi = require('ffi-napi');
+import path = require('path');
+import { FfiBinding } from './types';
+
+// This is a lookup between process.platform and
+// the platform names used in pact-reference
+const PLATFORM_LOOKUP = {
+  linux: 'linux',
+  darwin: 'osx',
+  win32: 'windows', // yes, 'win32' is what process.platform returns on windows 64 bit
```
mefellows, line 11: 😆

mefellows (approved):
Looks great Tim!

Couple of things:

Did you mean to commit the FFI deps? I was assuming we would package these at release step and perhaps have a developer step that would pull these down?

I'm just going to have to assume those hairy type files work, they look complicated, so they must be right, right? (also there are tests).

I actually like the idea of keeping the integration tests of the verifier. We may even want to extend them to ensure functional compatibility remains when we upgrade.

src/ffi/internals/index.ts
TimothyJones: It's "correct", because the name of the windows API layer is "win32". Which is a terrible name 😂

src/logger.ts
TimothyJones: It was a good idea that one @mefellows had :)

src/verifier/nativeVerifier.ts
TimothyJones: Right, I'll make it set and pass down the env variable.

src/verifier/types.ts
```diff
@@ -0,0 +1,33 @@
...
+  pactBrokerToken?: string;
+  consumerVersionTags?: string | string[];
+  providerVersionTags?: string | string[];
+  providerStatesSetupUrl?: string;
+  publishVerificationResult?: boolean;
+  providerVersion?: string;
+  enablePending?: boolean;
+  // Todo in FFI
+  includeWipPactsSince?: string;
+  consumerVersionSelectors?: ConsumerVersionSelector[];
```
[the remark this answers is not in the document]
TimothyJones: Ah, I've fixed both issues.

src/verifier/types.ts
TimothyJones: I actually prefer options 1 and/or 2 - if something explodes mid run, we've got the best chance to have the logs. And, if the pattern is to defer logging to the core, then we'll get more consistency across pact projects in general.

test/verifier.integration.spec.ts
TimothyJones: No, it's not here for that. I left it mostly because monkeypatch shouldn't make anything explode (it just warns now). But I think it's pretty low value. I'd be happy to have it removed.

TimothyJones: (especially as `monkeypatch.rb` is not where this test says it is)

src/verifier/types.ts
TimothyJones: Whichever option we choose, I think they are all clear arguments for a single ffi lib.

test/verifier.integration.spec.ts
TimothyJones: I removed monkeypatch.rb, but left the test. Not completely convinced it's necessary, though

src/verifier/types.ts
mefellows: I agree on the single FFI lib. That will take a bit of work, but not massive amounts. 

In the meantime, the log functions are all exposed in the mock server lib, which will apparently (untested by me) work as if it were a single application anyway. 

Good point on the crash and consistency. FWIW it's what I'm doing with Go. I have the output in the `go test` looking pretty nice now, so a separate log file or just enabling a more verbose log level is probably going to be just fine.

The log function I originally proposed lets you log the context across the boundary e.g. something like `pact-js-core:<package>:<subpackage>` could prefix all log entries, so you could quickly identify where to debug next.

mefellows: I'm actually wondering if we remove this property altogether, and just have a global logger interface that lets us set the level. Then it can be independent of the verifier and the mock server - what do you think?

TimothyJones: I think most people are likely to use the constructors or factories as entry points into pact - and it's convenient for those to have a `logLevel` property. Even though we expose multiple tools, typically people are only using one of them in a given context, I think. But, I do think that we can do better with config, in general. The config proposal is related, here: https://github.com/pact-foundation/pact-js/issues/602

src/logger.ts
```diff
@@ -3,21 +3,48 @@ import pino = require('pino');
...
+
+This is almost certainly a bug in pact-js-core. It would be great if you could
+open a bug report at: https://github.com/pact-foundation/pact-js-core/issues
+so that we can fix it.
+
+There is additional debugging information above. If you open a bug report, 
+please rerun with logLevel: 'debug' set in the VerifierOptions, and include the
+full output.
+
+SECURITY WARNING: Before including your log in the issue tracker, make sure you
```
bethesque, line 38: 😆

src/verifier/arguments.ts
```diff
@@ -0,0 +1,97 @@
...
+          return [...acc, '--file', uri];
+        default:
+          return acc;
+      }
+    }, []),
+  pactBrokerUrl: { arg: '--broker-url', mapper: 'string' },
+  pactBrokerUsername: { arg: '--user', mapper: 'string' },
+  pactBrokerPassword: { arg: '--password', mapper: 'string' },
+  pactBrokerToken: { arg: '--broker-token', mapper: 'string' },
+  consumerVersionTags: (tags: string | string[]) => [
```
bethesque, line 46: Does the rust impl not support consumer version selectors yet?

bethesque: Oh, don't worry, I found it below.

TimothyJones:
Closing this as it's obviated by the already merged version

PR #302, opened 2021-07-05
feat: improve m1 support (via rosetta2)

johanneswuerbach:
Related to https://github.com/pact-foundation/pact-js-core/issues/264

Instead of forcing the entire node environment to be executed using rosetta2, which requires manual work, install the existing binaries and rely on the fact that macOS is automatically using rosetta2 to execute them. This already works by simply removing the `arm64` restriction.

While https://github.com/pact-foundation/pact-js-core/pull/271 was closed, I'm not sure how realistic native M1 binaries are in the mid-term looking at the fact that traveling-ruby (which is used for the current binaries) has no ARM support https://github.com/phusion/traveling-ruby/issues/55 and the latest version is also not a simple update due to missing 32bit support https://github.com/phusion/traveling-ruby/issues/88#issuecomment-797509333  

On an M1 this now allows to run `npm install` with a default node version and to run our contract tests without any manual work, which would be a really good development experience improvement for us.


TimothyJones:
Can you elaborate on why this is the appropriate fix? My understanding from the package json documentation here:

https://docs.npmjs.com/cli/v7/configuring-npm/package-json#cpu

is that the cpu key is for listing the architecture that the code runs on. As the ruby core doesn’t support arm, my understanding is that it’s not appropriate to fix it this way. Do you have different information?

We are working on using ffi bindings to the Rust core, which might make the path to supporting M1 easier. I’m keen to know if the pact binaries in pact-reference can be built on an M1 mac.

There’s also a CI hurdle- GitHub actions don’t currently support M1 builds (as far as I can tell). Although that is probably just a matter of time. I’ve wondered about doing the builds manually until we can do that, but I don’t yet have an M1 machine either.

TimothyJones:
Because (unless I am missing something) this compromises the meaning of the `cpu` field, I'm not super keen to see this merged in to the mainline. 

But, if there really is no other workaround at the moment, perhaps we can release this as a one off pre-release with a warning on it, just so you can be unblocked.

johanneswuerbach:
> As the ruby core doesn’t support arm

I think in this case it is a bit more tricky. While the ruby core isn't natively working on an M1, it is automatically translated using rosetta 2 and appears to be working. Compared the other workaround the main difference is that the package can now be installed from node version running as arm64, while without this change node also needs to be run via rosetta 2.

So the main difference is imo whether you ask users to translate nodejs & pact (which requires manual work) or whether macos is only translating pact, which it does automatically.

I'm happy to add a warning or similar, if this situation is detected and a note to the README, if you think this approach makes sense.

Will look into compile the rust bindings, I haven't seen that there is WIP moving away from ruby.

johanneswuerbach:
The rust bindings compile for me, is there a list of work left to be done or timeline somewhere available when those are going to replace the ruby implementation?

Nevertheless I still think this PR improves the developer experience today as we see people struggle to install pact on their M1 machines.

TimothyJones:
Yeah. I’ve been thinking about this a bit, and I think accepting the risk of being wrong in the limited number of cases where people are on that architecture and don’t have Rosetta is probably worth saying we support it until we do.

In other news, I have ordered an M1 laptop (in part) so I can take a look more directly.

TimothyJones:
>  is there a list of work left to be done or timeline somewhere available when those are going to replace the ruby implementation?

Yes-ish. Currently, this is the next step, https://github.com/pact-foundation/pact-js/issues/675 which would be a contribution to this branch: https://github.com/pact-foundation/pact-js-core/tree/beta

If you're interested in helping out directly, shoot me a message in the pact slack and I'll take you through the progress so far. I'm hoping to work on it this weekend, but then I also hoped that for the last two weekends.

TimothyJones:
So, I'm still uncomfortable with claiming that we support `arm64` in the package.json before we do, but pragmatically I don't think we'll have many users who are using `arm64` without rosetta available. I reckon we use this approach for now. Thanks for the PR and discussion, @johanneswuerbach, and apologies for taking so long on this one!

TimothyJones:
pact-node@10.13.1 / pact-core@11.1.0 are now released with this workaround.

Issue #259, opened 2021-01-28
Remove request as a dependency


TimothyJones:
We have a dependency on https://www.npmjs.com/package/request , which is deprecated (and has been for some time).

To avoid everyone getting a warning, let's remove it.

I suggest replacing it with `axios`, but I have no strong opinion on it. I also haven't yet done a review of where we're using it.

mefellows:
Looks like we use it in the installer to download the ruby binaries dynamically, and in the service code to check if the Ruby Mock Service is up and running.

Axios is close enough to a default choice, and the project seems to have steered itself out of the wilderness, so is a safe enough choice I think.

adambocco:
Hello, 
I'd like to work on this issue. It looks like just the two places @mefellows mentions in src/service.ts and standalone/install.ts. 
Thanks

mefellows:
Amazing - thanks @adambocco .

I think @TimothyJones has made some progress/thinking on this. Tim - anything worth sharing?

If you're not already, please join slack.pact.io (and join the #pact-js-development channel) and we can chat there if you run into issues

TimothyJones:
Yes please, that would be excellent!

I took a stab at it before dinner yesterday, and found it a bit annoying to give proxy details to axios. It’s definitely possible, but might be more convenient to use a different library. I didn’t have anything worth saving- all I discovered is “this is more than 10 mins”.

Implementation wise and library choice, it’s up to you. I’d be happy to see any PR resolving the deprecation warnings

TimothyJones:
@adambocco Did you get anywhere with this? We're about to touch the installer code for a different reason, so if you have any work towards the solution it would be the perfect time to get it in.

PR #303, opened 2021-07-07
fix: Replace request with needle

Sertion:
I replaced [request](https://www.npmjs.com/package/request) with [needle](https://www.npmjs.com/package/needle). The [`requestCert` option used in service.ts](https://github.com/pact-foundation/pact-js-core/blob/6ca633864e380861e7da394c5a485c68aae17604/src/service.ts#L328) appears to only be used [when _starting a server_](https://nodejs.org/docs/latest-v14.x/api/tls.html#tls_tls_createserver_options_secureconnectionlistener) not when sending a request.


Sertion:
Should fix #259

TimothyJones:
This is awesome! Thank you so much. Code looks good, assuming the builds pass we'll merge this in.

TimothyJones (approved):
Nice work, thank you so much!

PR #315, opened 2021-08-23
chore: download all dependencies at compile time

mefellows:


script/download-ffi.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -eu
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)" # Figure out where the script is running
 . "$SCRIPT_DIR"/ci/lib/robust-bash.sh
-
+. "$SCRIPT_DIR"/ci/lib/utils.sh
```
TimothyJones, line 4: Could we call this something that isn't `utils`? In this case, it looks like `download-file.sh` might be alright

script/download-libs.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -eu
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)" # Figure out where the script is running
+
+${SCRIPT_DIR}/download-ffi.sh
```
TimothyJones, line 4: I think this needs to be quoted, otherwise it won't succeed if there's spaces in the path:

```
"${SCRIPT_DIR}"/download-ffi.sh
```

script/download-standalone.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash -eu
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)" # Figure out where the script is running
+. "$SCRIPT_DIR"/ci/lib/robust-bash.sh
+. "$SCRIPT_DIR"/ci/lib/utils.sh
+
+require_binary curl
+require_binary unzip
+
+STANDALONE_VERSION=$(cat ./standalone/install.ts | grep "PACT_STANDALONE_VERSION = '" | egrep -o "([0-9][\.0-9]+[0-9])")
```
TimothyJones, line 9: shellcheck will complain about the quotes here too, although it doesn't matter as much.

standalone/install.ts
```diff
@@ -1,38 +1,13 @@
-import * as http from 'http';
-import * as needle from 'needle';
-import unzipper = require('unzipper');
-import tar = require('tar');
-import pactEnvironment from '../src/pact-environment';
```
TimothyJones, line 5: 🙌

script/download-ffi.sh
mefellows: Why did I know you'd say that :P

script/download-libs.sh
mefellows: I just copied your use of it elsewhere - i'll fix all of them.

script/download-standalone.sh
```diff
@@ -0,0 +1,47 @@
...
+  STANDALONE_FILENAME="$2"
+
+  URL="${BASEURL}/v${STANDALONE_VERSION}/${1}"
+  DOWNLOAD_LOCATION="$STANDALONE_DIR/${STANDALONE_FILENAME}"
+  echo $URL
+
+  log "Downloading verifier $STANDALONE_VERSION to $DOWNLOAD_LOCATION"
+  download_to "$URL" "$DOWNLOAD_LOCATION"
+  if [ "${STANDALONE_FILENAME%zip}" != "${STANDALONE_FILENAME}" ]; then
+    unzip "$DOWNLOAD_LOCATION" -d "${DOWNLOAD_LOCATION%.*}"
```
TimothyJones, line 30: If we extract these when we download them, I think the npm package will be much larger. I was thinking it'd be better to extract them when they're needed instead

TimothyJones (commented):
This is great!

script/download-standalone.sh
TimothyJones: (I think the ffi doesn't matter as much, because it's only one file, and it's smaller. Although maybe it would be better to only extract as needed in general)

mefellows: The npm package is a tarball, so it will be compressed anyway.

script/download-standalone.sh
mefellows: Good reminder to re-install shell check, thanks. I removed a lot of extensions a while back because problems.

script/download-standalone.sh
mefellows: Just did a `--dry-run` locally to confirm this:

```
npm notice === Tarball Details ===
npm notice name:          @pact-foundation/pact-core
npm notice version:       12.0.0
npm notice package size:  29.9 MB
npm notice unpacked size: 87.0 MB
npm notice shasum:        cc0fb615407c4f6d0f0580196816e67877beb325
npm notice integrity:     sha512-zY8qGF8Qpg207[...]nMb1mVcRf7mIA==
npm notice total files:   2419
npm notice
```

You can see that the package is compressed, so no need to double compress it.

mefellows:
I think this is good to go Tim.

DEVELOPER.md
```diff
@@ -1,11 +1,8 @@
...
-No, you can't just run `npm it`, because that would be too easy:
-
-do this and you should be 👌👌👌:
+Do this and you should be 👌👌👌:
 
 ```
-npm ci --ignore-scripts
+npm ci
 npm test
-npm run download-checksums
```
TimothyJones, line 10: Woohoo!!

PR #316, opened 2021-08-25
feat: The new verifier now works on windows too

TimothyJones:
This was a bit of an ordeal. I refactored a bit along the way.

I think we should squash merge this one


src/pact-standalone.spec.ts
```diff
@@ -37,146 +37,154 @@ describe('Pact Standalone', function () {
   });
 
   describe('Check if OS specific files are there', () => {
-    describe('OSX', () => {
-      beforeEach(() => (pact = standalone('darwin')));
+    if (!process.env.ONLY_DOWNLOAD_PACT_FOR_WINDOWS) {
```
mefellows, line 40: 👌

TimothyJones:
Also the appveyor build is a red herring

script/lib/robust-bash.sh
```diff
@@ -10,11 +10,18 @@ if [ -z "${LIB_ROBUST_BASH_SH:-}" ]; then
     echo "🔵  ${1:-}"
   }
 
+  function debug_log {
```
mefellows, line 13: moar emoji!

mefellows (approved):
Awesome Tim. Skipping the bits on Windows that don't work makes sense to me.

We'll probably need some integration test of sorts at some point to ensure that the released artefact actually works on Windows. But, one problem at a time...

script/lib/robust-bash.sh
TimothyJones: 👀

src/pact-standalone.spec.ts
TimothyJones: Haha! I'd like not to have to do this. But it gets it going for now.

PR #323, opened 2021-09-24
feat: add latest ffi methods to declarations

mefellows:
Adding all of the FFI methods I've used in Pact Go, as well as some of the newer logging methods exposed.

I'm not sure about the Enums, but we're going to need them when we call the FFI functions in place of all of the `'int'` declarations. We can move them later, but I popped them in there now - thoughts?

Next step will be to expose a useful interface.

### DIFF
--- a/src/ffi/declarations.ts
+++ b/src/ffi/declarations.ts
@@ -30,12 +30,17 @@ export type FfiDeclarations = {
     [typeof PactHandle, 'string', 'bool']
   ];
   pactffi_new_pact: [typeof PactHandle, ['string', 'string']];
+  pactffi_with_specification: ['void', [typeof PactHandle, 'int']];
   pactffi_new_interaction: [
     typeof InteractionHandle,
     [typeof PactHandle, 'string']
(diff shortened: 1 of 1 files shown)


TimothyJones:
Looks good! Are the enums right? It might be best to not use typescript enums for them, possibly. I’m not sure.

mefellows:
> Looks good! Are the enums right? It might be best to not use typescript enums for them, possibly. I’m not sure.

Don't know yet, we'll find out when we go to use them. I just wanted to get in there what I had working elsewhere and ensure the correct mapping was documented. 

That being said, because they are numerical and map to the underlying Rust enum (that has been aligned correctly) we can always use the numerical value if not. 

The FFI signatures in JS are very hard to reason about i.e. the arguments ['string', 'int', 'int'...] aren't named, so it isn't very helpful to see what goes in there. I considered adding some type aliases, but perhaps adding the c interface as comments above each mapping will suffice as they will be wrapped by our own properly typed interface.

PR #327, opened 2021-09-30
Feat/ffi consumer

TimothyJones:
This adds a first cut of a consumer interface for the V3 / ffi bindings. There's a test that illustrates how to use it.


mefellows:
This is awesome 👏 . First glance it looks good, but let me dive deeper with fresh eyes and some time.

mefellows (approved):
Looks great, I'm going to pull this in - There needs to be a cheffkiss emoji !

There are a couple of methods missing from the exported interface that will be necessary when retiring the Rust portion of the code (`given_with_param`, `pactffi_with_multipart_file` and `pactffi_with_binary_file` at a glance) but i'll bring them in after.

TimothyJones:
We'll need a solution to the node 16 problem before we can release this.

mefellows:
Yeah I saw that. Albeit, Pact JS (v3) doesn't currently support Node 16 anyway? Should we temporarily disable Node16 builds here until we find a solution?

TimothyJones:
Currently it does support node 16, but not with musl.

TimothyJones:
Node 16 is the current release, I don't think we can release without it
https://nodejs.org/en/about/releases/

mefellows:
Ah, yes. Looks like an ffi rabbit hole will be adventured shortly...

PR #349, opened 2021-11-15
feat: support specifying provider branch in verifications

mefellows:
Fixes https://github.com/pact-foundation/pact-js/issues/750 (for the `pact-core` package)

### DIFF
--- a/src/ffi/index.ts
+++ b/src/ffi/index.ts
@@ -7,7 +7,7 @@ import { declarations, FfiDeclarations } from './declarations';
 
 const LOG_ENV_VAR_NAME = 'PACT_LOG_LEVEL';
 
-const PACT_FFI_VERSION = '0.0.3';
+const PACT_FFI_VERSION = '0.1.0';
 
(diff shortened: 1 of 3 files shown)


mefellows:
Sorry, hit review too early Tim. I need to upgrade to the latest FFI and test it out as per the discussion in the #pact-rust channel recently.

TimothyJones:
It's also worth diff'ing the C header file. I was thinking maybe we should check it in to make changes more obvious

TimothyJones:
What is the status of this one?

mefellows:
Ah, the status is that bumping the FFI version alone won't work, despite the flag being correct.

The issue is that the FFI mappings need to change a little with the newer n

See https://pact-foundation.slack.com/archives/CA2S7E6KC/p1637046326013500 and https://pact-foundation.slack.com/archives/CA2S7E6KC/p1637047542017900 for some context.

TL;DR - another version bump is required and then the two structs can just be referenced as opaque pointers (the implementation details are now hidden from us).

mefellows:
`providerBranch` replaces this in current master. Closing defunct PR.

PR #350, opened 2021-11-15
feat: support specifying provider branch in verifications

mefellows:
Fixes https://github.com/pact-foundation/pact-js/issues/750 (for the `pact-node` package)


TimothyJones:
Nice. I think it would be ideal if users didn’t have to configure this, though

mefellows:
Yeah. There is the option on the consumer side, so it would make sense to have a similar auto-detect on the provider.

PR #356, opened 2021-12-15
fix: Serialize {enablePending: false} to --no-enable-pending

flappyBug:
This should fix https://github.com/pact-foundation/pact-js/issues/782

### DIFF
--- a/src/spawn/arguments.ts
+++ b/src/spawn/arguments.ts
@@ -13,6 +13,13 @@ export type CliVerbOptions = {
   cliVerb: string;
 };
 
+export interface ArgumentMappings {
+  [key: string]: Mapping;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Mapping = string | ((val: any) => string[]);
+
 export type SpawnArgument =
   | CanDeployOptions
   | MessageOptions
@@ -34,11 +41,12 @@ const valFor = (v: SpawnArgument): Array<string> => {
   return v !== PACT_NODE_NO_VALUE ? [`${v}`] : [];
 };
 
-const mapFor = (mapping: string, v: string): Array<string> =>
-  mapping === DEFAULT_ARG ? valFor(v) : [mapping].concat(valFor(v));
+const mapFor = (mapping: Mapping, v: string): Array<string> =>
+  mapping === DEFAULT_ARG ? valFor(v) :
+    typeof mapping === 'string' ? [mapping].concat(valFor(v)): mapping(v);
 
 const convertValue = (
-  mapping: string,
+  mapping: Mapping,
   v: SpawnArgument | Array<SpawnArgument>
 ): Array<string> => {
   if (mapping && (v || typeof v === 'boolean')) {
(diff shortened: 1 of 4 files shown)


src/spawn/spawn.ts
```diff
@@ -5,6 +5,7 @@ import * as path from 'path';
 import logger from '../logger';
 import pactEnvironment from '../pact-environment';
 import argsHelper, { SpawnArguments, DEFAULT_ARG } from './arguments';
+import { ArgumentMappings } from '.';
```
TimothyJones, line 8: I think this is meant to be `./arguments` ?

src/spawn/arguments.ts
```diff
@@ -13,6 +13,13 @@ export type CliVerbOptions = {
   cliVerb: string;
 };
 
+export interface ArgumentMappings {
+  [key: string]: Mapping;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Mapping = string | ((val: any) => string[]);
```
TimothyJones, line 21: I'd prefer not to turn off the lint warnings - perhaps this could be `unknown`?

TimothyJones (commented):
Thanks very much for this! Two small issues:

1) The commit messages with `fix` or `feat` appear in the changelog - would you be able to either:
--  force push a new message that explains the change for the changelog (something like "correct behaviour of enablePending: false in the verifier API")
--  let us know that you're ok with a squash merge to rewrite the message

2) I've left a couple of review comments inline - the one I care about the most is the lint directives - we usually only allow `any` if a dependency is requiring it

TimothyJones:
I forgot to mention in the review above: This is nice work, thank you!

src/spawn/arguments.ts
flappyBug: When we change to `unknown` here, we need annotate the mapping function argument type with unknown,
```ts
let argMapping: ArgumentMappings = {
    enablePending: (enablePending: unknown) => [enablePending ? '--enable-pending' : '--no-enable-pending']
}
```
or do a explicit cast:
```ts
let argMapping = {
    enablePending: (enablePending: boolean) => [enablePending ? '--enable-pending' : '--no-enable-pending']
} as ArgumentMappings;
```
Is that ok?

TimothyJones: Hm, I'd rather not do an explicit cast unless we have to, as (like `any`) it's turning off the type checking. Maybe we can bring in an implicit type parameter or something - I'll have a play.

TimothyJones: Ah, right - so the type checking is already pretty nobbled by this point, because of the way the rest of the mapper works (there are also already some explicit casts in the branches above). This is one of the reasons we rewrote the mapper for the native verifier in the master branch. 

I think `unknown` is probably the better way-  I'll push a commit that does that.

TimothyJones: (I prefer `unknown`, because really we don't know what the type will be at that point. If we did rewrite this mapper from scratch, we'd want to do it in a way that the type information was retained. I think this is an artefact of this originally being pure JS)

TimothyJones:
Thanks again for the PR - making it super easy to get this fixed. 10.17.0 is now released.

PR #359, opened 2022-02-02
feat: collect usage telemetry data

Inksprout:
This PR adds two environment variables, PACT_EXECUTING_LANGUAGE and PACT_EXECUTING_LANGUAGE_VERSION. These variables will be used to track metrics such as when consumer and provider tests are ran so that we(Pact OSS maintainers) can gain insight into what languages and versions are used for Pact. PRs for some of the metrics [here ](https://github.com/pact-foundation/pact-mock_service/pull/133) and [here](https://github.com/pact-foundation/pact-ruby/pull/256)

### DIFF
--- a/src/spawn/spawn.ts
+++ b/src/spawn/spawn.ts
@@ -17,6 +17,10 @@ export class Spawn {
     argMapping: { [id: string]: string } = {}
   ): ChildProcess {
     const envVars = JSON.parse(JSON.stringify(process.env)); // Create copy of environment variables
+
+    envVars['PACT_EXECUTING_LANGUAGE'] = 'node.js'
+    envVars['PACT_EXECUTING_LANGUAGE_VERSION'] = process.versions.node
+
     // Remove environment variable if there
     // This is a hack to prevent some weird Travelling Ruby behaviour with Gems
     // https://github.com/pact-foundation/pact-mock-service-npm/issues/16
--- a/src/spawn/spawn.ts
+++ b/src/spawn/spawn.ts
@@ -18,8 +18,8 @@ export class Spawn {
   ): ChildProcess {
     const envVars = JSON.parse(JSON.stringify(process.env)); // Create copy of environment variables
 
-    envVars['PACT_EXECUTING_LANGUAGE'] = 'node.js'
-    envVars['PACT_EXECUTING_LANGUAGE_VERSION'] = process.versions.node
+    envVars['PACT_EXECUTING_LANGUAGE'] = 'node.js';
+    envVars['PACT_EXECUTING_LANGUAGE_VERSION'] = process.versions.node;
 
     // Remove environment variable if there
     // This is a hack to prevent some weird Travelling Ruby behaviour with Gems


TimothyJones:
I'm uncomfortable with this change. Previously we only used to track download statistics, updating pact to call home with more information seems strictly worse to me.

Also, who is "we"?

mefellows:
Hey Tim,

For general background about the why, what and how, see this [doc](https://docs.google.com/document/d/1F5kkfhFu7n74T8VKJ3NWWFhnr1WLndDHcBNirCeutIA/edit#) shared with the maintainers. We discussed making these analytics public, but initially it's envisaged the core maintainers would have access. See also the slack [thread](https://pact-foundation.slack.com/archives/G9VHV9VJ9/p1629171327042300).

> As above, I don't think we should accept this, though. At the very least, we need more information on who "we" is, and the use case.

"we" = the Pact OSS maintainers. The use cases are documented in the proposal, but essentially having usage metrics is helpful for the same reason that having analytics your website is helpful. Knowing how many people use what language and features helps us prioritise (e.g. should we do the Ruby thing first or the Golang one? What language is not well adopted, how can we change that etc.). 

> Tracking that a download happened is one thing, but tracking versions of runtimes is a whole different thing - as it's information exposure that would be useful to an attacker. And, since any information captured by the OSS tooling should be public, I think we should reject this change.

The IP data sent is [anonymised](https://support.google.com/analytics/answer/2763052?hl=en) (as per the previous install data). We don't know anything about the _user_, but we get a little bit about what tools they are using (in this case, that the user is on version X of node). Again, this information is good to know - a bit like the os/arch support.

I don't think this presents any security risk over and above what we previously did.

> @mefellows, what do you think? It's kind of my last day as maintainer, but I'd be really sad if immediately after leaving the project we updated it to call home more often. I think it's already pretty questionable that our tracking message is essentially "we just tracked you! you can disable this for next time if you want". That's not how consent for tracking should work.

It's pretty standard across the industry - for example both [AWS SAM CLI](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-telemetry.html) and [dotnet core](https://aka.ms/dotnet-cli-telemetry) emit telemetry in exactly the same way (we have "borrowed" from their messages). If they can do it, I can't see why we can't.

I think you mentioned GDPR in a related comment, and because we don't have any PII information we are compliant. Nothing in the data can be used identify the user. we also discuss relevant regulations in the document.

Overall, the collection of this data presents an opportunity to help better understand our audience and how we can better serve them.

Inksprout:
Hi Tim, 

Thanks for catching the issue with the types, my mistake! I have removed those changes and also updated the commit so that this change will appear in the New Features section of the release notes. I'm new to contributing to OSS so thank you for your patience. As for the other feedback see Matt's comments above.🙂  I also added to the PR description that the 'we' in question are the pact OSS maintainers as Matt mentioned.

TimothyJones:
>  see this doc shared with the maintainers

I remember reading this at the time, and my impression (from admittedly a skim read) was that this was just codifying the tracking behaviour that is part of pact-js, with the view to moving it into the core. I didn't realise it was extending the tracking to cover things that are not part of the download statistics (the proposal actually says "we already do [collect metrics] in pact-node").

The part of the proposal that extends what is collected is only in the detail at the end. I know you weren't hiding it on purpose, and if you do think it's a minor change then I see why it didn't make it to the executive summary at the top - but, I think this is a substantial change, and given the large amount of controversy tracking generally attracts, I'm sure I won't be the only one who thinks this is a substantial change.

> See also the slack thread.

The slack thread is about the proposal, and we've never reached a conclusion. I think we need to have a process where we confirm with the whole team before deciding that a proposal is ready to be implemented - even a simple "hey, this proposal has been around for 6 months, anyone see a reason we can't make it a reality?" would have covered this. Certainly that would trigger a close read of the proposal. 

> If they can do it, I can't see why we can't.

I think this is about surprise:

* It is uncommon, but not surprising that an npm package does extra downloads during install
* It is not surprising for download statistics to be tracked on the server. Because we don't control the server in this case, we do that tracking separately.
* It *is* surprising that extra information about the user's system and architecture is tracked as part of this download tracking.

>  'we' in question are the pact OSS maintainers as Matt mentioned.

"We" the Pact OSS maintainers are not in agreement on this feature, and I think there is a need for further discussion before it is merged.


> Because we don't have any PII information we are compliant (or rather, not in scope).

Although I'm not an expert on the subject, I don't think this is true. My understanding is that the GDPR is explicit in that it covers "information that can be used to identify the user". This would cover the collection of system statistics that are unrelated to the operation of the product. If we are collecting those, then my understanding is that we must be opt-in.

Even if most of the maintainers think the extension to non-download related statistics is trivial, I think:

* Tracking that is not required as part of operation exposes us to the "must be opt-in" clause of the GDPR. We should definitely confirm this before release, as the penalties are substantial
* Even ignoring the law, general tracking culture is opt-in now, and having non-download tracking be opt out risks exposing the project to bad press

TimothyJones:
> Thanks for catching the issue with the types, my mistake! I have removed those changes and also updated the commit so that this change will appear in the New Features section of the release notes.

Thanks @Inksprout ! Note that what is appropriate to appear in the changelog is potentially different to a regular commit message. For example, at the moment, the changelog would read:

* collect usage telemetry data

A better message might be something like: 

* Pact now collects platform version information as well as download numbers. For a full discussion of what we collect, see (link)

(I think if we do merge this, it would be good to have a link to the documentation - but I think that documentation has not been written yet).

mefellows:
I like the issue SAM CLI created to announce what they were doing: https://github.com/aws/aws-sam-cli/issues/1272. Perhaps we should template up an issue like this, and if we roll it out across the project use the same one so there is a record in each repo of the change? 

> Although I'm not an expert on the subject, I don't think this is true. My understanding is that the GDPR is explicit in that it covers "information that can be used to identify the user". This would cover the collection of system statistics that are unrelated to the operation of the product. If we are collecting those, then my understanding is that we must be opt-in.

GDPR is all about protecting personal data. If we don't collect anything that can be used to identify a user, it's out of scope. It doesn't mean we shouldn't protect it, it just means that regulation is not applicable.

From [article 4](https://gdpr.eu/article-4-definitions/) of the regulation:

> ‘personal data’ means any information relating to an identified or identifiable natural person (‘data subject’); an identifiable natural person is one who can be identified, directly or indirectly, in particular by reference to an identifier such as a name, an identification number, location data, an online identifier or to one or more factors specific to the physical, physiological, genetic, mental, economic, cultural or social identity of that natural person;

None of the data we collect can be used to identify a user, and anything remotely close to that (IP) is hashed/anonymised.

I'm comfortable with the legal basis of collecting the information, although yes, I'm not a lawyer.

If it helps, both [AWS](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-telemetry.html#serverless-sam-telemtry-data-collected) and [dotnet](https://docs.microsoft.com/en-gb/dotnet/core/tools/telemetry#data-points) track a lot more data than what we are proposing here without explicit consent. Given this case, it's safe to assume they and their million strong legal team have endorsed the approach.

> (I think if we do merge this, it would be good to have a link to the documentation - but I think that documentation has not been written yet).

You're right, we should have a page about this on docs.pact.io (and link to it in the notice). I'm happy to coordinate that.

> The slack thread is about the proposal, and we've never reached a conclusion. I think we need to have a process where we confirm with the whole team before deciding that a proposal is ready to be implemented - even a simple "hey, this proposal has been around for 6 months, anyone see a reason we can't make it a reality?" would have covered this. Certainly that would trigger a close read of the proposal.

Yes, I think we have some work to do on process!

PR #386, opened 2022-08-09
Expose pactflow command

### LINKED ISSUE #385: expose pactflow command in root node_modules/.bin
As per https://github.com/pactflow/example-bi-directional-provider-dredd/issues/17

pact-js-core contains the latest pact-ruby-standalone which contains the pactflow command.

It will be available in the following path 

node_modules/@pact-foundation/pact-core/standalone/linux-x64-1.89.02-rc1/pact/bin/pactflow

but is not mapped to the root `./node_modules/bin` folder.

It needs two things

1. A pactflow.js file adding to https://github.com/pact-foundation/pact-js-core/tree/master/bin
2. An entry for the pactflow command here https://github.com/pact-foundation/pact-js-core/blob/master/package.json#L14

YOU54F:
closes #385

### DIFF
--- a/bin/pactflow.ts
+++ b/bin/pactflow.ts
@@ -1,0 +1,13 @@
+#!/usr/bin/env node
+
+import standalone from '../src/pact-standalone';
+import childProcess = require('child_process');
+
+const status = childProcess.spawnSync(
+  standalone.pactFullPath,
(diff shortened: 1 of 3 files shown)


mefellows:
Ah, thank you!

PR #393, opened 2022-08-12
fix: corrects customProviderHeaders validation

adamwitko:
### PR Template

This PR fixes the changes made to validation in 13.7.3 which did not support the types specified for customProviderHeaders.  

Fixes (pact-foundation/pact-js-core#392)


mefellows:
Thanks Adam. I'm going to also update this code to allow the backwards compatible `string[]` type as well. But let's get this in for now.

PR #408, opened 2022-09-28
feat: support branch/autoDetectVersionProperties when publishing pacts

YOU54F:
Addresses https://github.com/pact-foundation/pact-js/issues/749

Haven't added any tests in, but have tested with a local broker instance (spun up from the pact-workshop-js) repo

if `autoDetectVersionProperties` is true, but branch is set, it will use the provided `branch` value.
if `autoDetectVersionProperties` is true, but branch is not set, it will pick up the branch from my local machine
if `branch` is set, and is an empty string or undefined, no branch is set when uploading
if `branch` is set, is a valid string and `autoDetectVersionProperties` is not set to true, the user provided value is set

```
var pact = require('./src/pact');
var path = require('path');
const { Pact } = require('./src/pact');

var PactInstance = new Pact();

PactInstance.publishPacts({
  pactBroker: 'http://localhost:8000',
  pactBrokerUsername: 'pact_workshop',
  pactBrokerPassword: 'pact_workshop',
  pactFilesOrDirs: [
    path.resolve(__dirname, 'FrontendWebsite-ProductService.json'),
  ],
  consumerVersion: '1.0.3',
  branch: 'foo',
  autoDetectVersionProperties: true,
}).then(function (res) {
  console.log(res);
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,12 +55,12 @@ In order to get better statistics as to who is using Pact, we have an anonymous 
 
 ```json
 {
-	"name": "some-project",
-	...
-	"config": {
-		"pact_do_not_track": true
-	},
-	...
+ "name": "some-project",
+ ...
+ "config": {
+  "pact_do_not_track": true
+ },
+ ...
 }
 ```
 
@@ -87,7 +87,7 @@ Simply require the library and call the create function to start the mock servic
 var pact = require("@pact-foundation/pact-core");
 var server = pact.createServer({ port: 9999 });
 server.start().then(function() {
-	// Do your testing/development here
+ // Do your testing/development here
 });
 ```
 
@@ -97,7 +97,7 @@ Or if you're using Typescript instead of plain old Javascript
 import pact from "@pact-foundation/pact-core";
 const server = pact.createServer({ port: 9999 });
 server.start().then(() => {
-	// Do your testing/development here
+ // Do your testing/development here
 });
 ```
 
@@ -127,7 +127,7 @@ Mock servers are used by Pact to record interactions and create pact contracts.
 ```js
 var pact = require('@pact-foundation/pact-core');
 var server = pact.createServer({
-	...
+ ...
 });
 ```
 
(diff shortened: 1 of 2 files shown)


bethesque:
Is autoDetectVersionProperties working as you would expect?

YOU54F:
> Is autoDetectVersionProperties working as you would expect?

Well, it works how I expect it to behave, based on how I have interacted with the pact_broker-client, however that isn't to say, that is how one would possible expect it to behave.

1. Based on the output of setting the parameter `autoDetectVersionProperties`, we will automatically detect the branch. It feels like `autoDetectBranchProperty` would be more apt/intiative
2. As a user provided branch value takes precedence over the computed branch value determined if `autoDetectVersionProperties` is set to `true`, could we enable branch detection (and application) automatically, with the user having to opt out, or manually set a `branch` property? Rationale below
  - We want to get users to publish with branch set so they can use the new webhook event `contract_requiring_verification_published`
  - Users can override it being set, by passing the branch property 
  - Users could opt out completely
  - It should cover the majority of use cases (is there a case where someone wouldn't want to publish with a branch )
 3. Is the ideal world of compatibility with our old docs/badges etc to advise publishing with branch and a tag representing the branch, or what would the potential implications be in applying a tag with the name on the branch automatically)
   - `--auto-detect-version-properties --tag-with-git-branch`

bethesque:
I think I meant the autoDetectVersionProperties to also do the version number, and the build URL, and just hadn't gotten around to doing them.

I wanted it to default to true, so it would just work without people having to do anything.

> Is the ideal world of compatibility with our old docs/badges etc to advise publishing with branch and a tag representing the branch, or what would the potential implications be in applying a tag with the name on the branch automatically)

Nothing terrible would happen - it's just a bit messy. The Pact Broker already takes the first tag, and turns it into the branch name (unless someone disabled this config) to allow people to easily migrate.

bethesque:
> I think I meant the autoDetectVersionProperties to also do the version number, and the build URL, and just hadn't gotten around to doing them.

Should I add in the version from the git sha?

YOU54F:
That sounds sensible and looks like it was a feature req too

https://github.com/pact-foundation/pact_broker-client/issues/100

PR #411, opened 2022-10-14
feat: expose faiIfNoPactsFound on the VerifierOptions

adamwitko:
This PR is to add in the ability for Providers to specify if they want to pass or fail when no pacts can be found to verify.

Fixes (pact-foundation/pact-js#941)

### DIFF
--- a/src/verifier/argumentMapper/arguments.ts
+++ b/src/verifier/argumentMapper/arguments.ts
@@ -31,9 +31,10 @@ export const orderOfExecution: OrderedExecution = {
   pactffiVerifierSetVerificationOptions: 4,
   pactffiVerifierSetPublishOptions: 5,
   pactffiVerifierSetConsumerFilters: 6,
-  pactffiVerifierAddCustomHeader: 7,
-  pactffiVerifierAddDirectorySource: 8,
-  pactffiVerifierBrokerSourceWithSelectors: 9,
+  pactffiVerifierSetFailIfNoPactsFound: 7,
+  pactffiVerifierAddCustomHeader: 8,
+  pactffiVerifierAddDirectorySource: 9,
+  pactffiVerifierBrokerSourceWithSelectors: 10,
 };
 
 export type MergedFfiSourceFunctions = {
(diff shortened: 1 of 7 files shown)


adamwitko:
I am having issue with `make` on `npm run native` so it's proving a little difficult to add and run the tests before I push.

adamwitko:
- [ ] Look in to adding more tests as these changes feel quite underdocumented

mefellows:
Thanks @adamwitko. I've been off for a long weekend, will review this week - appreciate the PR.

Most of this work is just mapping arguments to the underlying process, where the real logic/behaviour exists. I'm less concerned about an integration test for this behaviour at this layer, but we could look to sure up the unit test coverage.

A lot of the mapping is enforced at the type (TS) level, but of course this doesn't guarantee behaviour/validation in JS land.

src/verifier/argumentMapper/arguments.ts
```diff
@@ -177,6 +178,18 @@ export const ffiFnMapping: FnMapping<
       return { status: FnValidationStatus.IGNORE };
     },
   },
+  pactffiVerifierSetFailIfNoPactsFound: {
+    validateAndExecute(ffi, handle, options) {
+      if (options.failIfNoPactsFound) {
```
mefellows, line 183: The main issue is that now the default behaviour is to _not_ fail if no pacts are found (which is not currently the case, and may be surprising to some users). I'll have to have a think if this is sensible or not, or if we should check if the value is defined and set the value to false.

We may need to consider adding the default somewhere higher up in the stack (e.g. Pact JS)

adamwitko: Late Friday mistake on my behalf then. I am more than happy to rework this as you are 100% right. This can pass in true if not provided else the value provided. I'm away from my laptop for the next few days but I can look at fixing in when I'm back

mefellows: All good. I'll add the check now and bring it in - much appreciated!

PR #419, opened 2022-11-30
fix: Update pact-ruby-standalone to 1.91.0, shortening the path for windows users

TimothyJones:
I'm not sure if pact-node is still being released, but this addresses [the regression](https://pact-foundation.slack.com/archives/C9VPNUJR2/p1669678385756549?thread_ts=1669658751.887949&cid=C9VPNUJR2) introduced in 10.17.3 - by updating the version to the main release instead of release candidate releases (which have longer version numbers, and are more likely to hit the long path issue)


mefellows:
Thanks 10.17.7 released 🎉

PR #425, opened 2022-12-22
Match pact js lint

TimothyJones:
This PR is like https://github.com/pact-foundation/pact-js/pull/1027 , except for pact-js-core.

Unfortunately, it's not as well-separated as the previous one. Some key points:

* The windows build wasn't failing if the build failed. I fixed this.
* Most of the classes modified the options object they were given (which is probably unexpected behaviour for the caller). I fixed this, but it was actually relied on by the service classes so I had to un-fix it. The tests for those classes weren't testing anything, because if you modify the input config object, then `expect(instance.options.foo).to.equal(options.foo)` is comparing the object with itself. 
    * It's possible that pact-js relies on this behaviour too 
* I've revamped the typescript settings to be much improved
* I fixed a problem where vscode wouldn't understand the build, because it can't read `tsconfig.spec.ts`. Unfortunately, this means that the mocha types need to be available to typescript during the main build - although the spec and test files aren't outputted.
* In `src/ffi/types.ts`, the recommended lint settings complain about my pattern of merging types and constants (the pattern exists to get around some enum unpleasantness). There's no nice way to whitelist this type of redeclaring, and the docs for the [no-redeclare](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-redeclare.md) rule say this is intentional. Probably the right fix is to use an enum instead :(
* Many files had their internal methods used before they were defined
* I split some files where the linter was complaining about too many classes
* I split out `types.ts` to prevent circular dependencies. The pattern is - 
   * exported types are in `types.ts` and not in any other files
   * types can only be imported internally from the relevant types.ts
   * the main index.ts exports all `types.ts` files. No internal code (except tests) imports from the root (otherwise you get circular dependencies).


mefellows (approved):
🎉

Issue #426, opened 2022-12-27
pact.publishPacts(opts) missing


ireinhart:
Since 13.3.3 the function publishPacts() is not found.

### Software versions

- **OS**: node:lts-bullseye-slim
- **Pact Node version**:  pact-foundation/pact@10.4.0
- **Node Version**: node:lts-bullseye-slim
- **Other Versions**: @pact-foundation/pact-core@13.13.3

### Issue Checklist

Please confirm the following:

- [x] I have upgraded to the latest
- [x] I have the read the FAQs in the Readme
- [x] I have triple checked, that there are **no unhandled promises** in my code
- [ ] I have set my log level to debug and attached a log file showing the complete request/response cycle
- [ ] For bonus points and virtual high fives, I have created a reproduceable git repository (see below) to illustrate the problem

### Expected behaviour

If I run the code from https://github.com/pact-foundation/pact-js-core#pact-broker-publishing it should publish pacts.

My version
```js
let publisher = require("@pact-foundation/pact-core");
let path = require("path");
if(process.env.bamboo_planRepository_branch == 'master') {
    let opts = {
        pactFilesOrDirs: [path.resolve(process.cwd(), "../../artifacts/contracts/js")],
        pactBroker: "https://pact-broker.mydomain.net",
        consumerVersion: "1.0-" + process.env.bamboo_repository_revision_number.substring(0, 8),
        tags: ['master']
    };
    publisher.publishPacts(opts);
} else {
    console.log('Pact publishing only at master branch on bamboo');
}
```

### Actual behaviour

```
build	27-Dec-2022 05:13:37	> app@1.0.0 publish:pact
build	27-Dec-2022 05:13:37	> node ./assets/js/pact/publish.js
build	27-Dec-2022 05:13:37	
error	27-Dec-2022 05:13:37	/opt/bamboo/xml-data/build-dir/APPLICATIONS-UM-JOB1/components/app/assets/js/pact/publish.js:10
error	27-Dec-2022 05:13:37	    publisher.publishPacts(opts);
error	27-Dec-2022 05:13:37	              ^
error	27-Dec-2022 05:13:37	
error	27-Dec-2022 05:13:37	TypeError: publisher.publishPacts is not a function
error	27-Dec-2022 05:13:37	    at Object.<anonymous> (/opt/bamboo/xml-data/build-dir/APPLICATIONS-UM-JOB1/components/app/assets/js/pact/publish.js:10:15)
[...]

PR #427, opened 2022-12-28
Fix cjs export

TimothyJones:
This PR fixes the regression reported in #426 by bringing back the export.

This rule should probably be brought back in the future, and this interface reviewed. At the very least, these functions should be explicitly exported instead of hacked through by exporting a class as the module export.

Also, I removed a mock that wasn't needed (and didn't compile anyway, since Jest isn't present)


TimothyJones:
And, apologies for the regression. Totally my fault.

YOU54F:
Thanks @TimothyJones, no problem, these things happen. Will review your comments in #426 and get a bit of a plan in action for some of the docs, will go through your other PR's today too

mefellows:
Thanks Tim!

> And, apologies for the regression. Totally my fault.

There's a first time for everything ;)

PR #428, opened 2023-01-09
feat: add pactffi_log_to_file for consumer/verifier

YOU54F:
# Add ability to log to file, via pactffi_log_to_file

Will go someway to address https://github.com/pact-foundation/pact-js/issues/954

(will need adding into pact-js)

Passed in via an optional param, to `getFfiLib` which is called by a couple of functions, as an optional param. Wonder if it would be better to have these in an opts object that can take named params.

note, the option used to be called `log` in pact-js v2 and is moved to deprecated now. should we retain `log` or use a more descriptive `logFile`

https://github.com/pact-foundaton/pact-js/tree/9.x.x#constructor

## edit

need to add some appropriate tests and refactor based on review comments :)

### DIFF
--- a/native/ffi.cc
+++ b/native/ffi.cc
@@ -49,6 +49,64 @@ Napi::Value PactffiInitWithLogLevel(const Napi::CallbackInfo& info) {
   return env.Undefined();
 }
 
+LevelFilter integerToLevelFilter(Napi::Env &env, uint32_t number) {
+  LevelFilter logLevel = LevelFilter::LevelFilter_Off;
+
+  switch(number) {
+    case 0:
+      logLevel = LevelFilter::LevelFilter_Off;
+      break;
+    case 1:
+      logLevel = LevelFilter::LevelFilter_Error;
+      break;
+    case 2:
+      logLevel = LevelFilter::LevelFilter_Warn;
+      break;
+    case 3:
+      logLevel = LevelFilter::LevelFilter_Info;
+      break;
+    case 4:
+      logLevel = LevelFilter::LevelFilter_Debug;
+      break;
+    case 5:
+      logLevel = LevelFilter::LevelFilter_Trace;
+      break;
+    default:
+      std::string err =  "pact-js-core C integration: Unable to parse log level number: ";
+      err += number;
+
+      throw Napi::Error::New(env, err);
+  }
+
+  return logLevel;
+}
+
+
+Napi::Value PactffiLogToFile(const Napi::CallbackInfo& info) {
+   // return: int
+   Napi::Env env = info.Env();
+
+  if (info.Length() < 2) {
+    throw Napi::Error::New(env, "PactffiLogToFile(envVar) received < 2 arguments");
+  }
+
+    if (!info[0].IsString()) {
+    throw Napi::Error::New(env, "PactffiLogToFile(envVar) expected a string");
+  }
+
+  if (!info[1].IsNumber()) {
+    throw Napi::Error::New(env, "PactffiLogToFile(envVar) expected a number");
+  }
+
+  // Extract log level
+  std::string fileName = info[0].As<Napi::String>().Utf8Value();
+  uint32_t levelFilterNumber = info[1].As<Napi::Number>().Uint32Value();
+  LevelFilter levelFilter = integerToLevelFilter(env, levelFilterNumber);
+
+  int res = pactffi_log_to_file(fileName.c_str(), levelFilter);
+
+  return Napi::Number::New(env, res);
+}
 /*
 
 Napi::Value Pactffi_log_message(const Napi::CallbackInfo& info) {
--- a/src/ffi/index.ts
+++ b/src/ffi/index.ts
@@ -2,7 +2,7 @@ import path from 'node:path';
 import bindings = require('node-gyp-build');
 import logger, { DEFAULT_LOG_LEVEL } from '../logger';
 import { LogLevel } from '../logger/types';
-import { Ffi } from './types';
+import { Ffi, FfiLogLevelFilter } from './types';
 
 export const PACT_FFI_VERSION = '0.4.22';
 
@@ -85,9 +85,8 @@ const renderBinaryErrorMessage = (error: unknown) => {
 };
 
 let ffi: typeof ffiLib;
-let ffiLogLevel: LogLevel;
 
-const initialiseFfi = (logLevel: LogLevel): typeof ffi => {
+const initialiseFfi = (): typeof ffi => {
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
   if (process.stdout._handle) {
@@ -95,8 +94,6 @@ const initialiseFfi = (logLevel: LogLevel): typeof ffi => {
     // @ts-ignore
     process.stdout._handle.setBlocking(true);
   }
-  logger.debug(`Initalising native core at log level '${logLevel}'`);
-  ffiLogLevel = logLevel;
   try {
     bindingPaths.every((bindingPath, i) => {
       try {
@@ -114,7 +111,6 @@ const initialiseFfi = (logLevel: LogLevel): typeof ffi => {
         return true;
       }
     });
-    ffiLib.pactffiInitWithLogLevel(logLevel);
   } catch (error) {
     renderBinaryErrorMessage(error);
     throw new Error(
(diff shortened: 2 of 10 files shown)


native/ffi.cc
```diff
@@ -49,6 +49,64 @@ Napi::Value PactffiInitWithLogLevel(const Napi::CallbackInfo& info) {
...
+      level = LevelFilter::LevelFilter_Info;
+      break;
+    case 4:
+      level = LevelFilter::LevelFilter_Debug;
+      break;
+    case 5:
+      level = LevelFilter::LevelFilter_Trace;
+      break;
+    default:
+      std::string err =  "Unable to parse Level Filter: ";
```
TimothyJones, line 75: This would be better if it included more about the source (eg, pact-js-core C integration, and "log level" instead of "Level")

src/ffi/index.ts
```diff
@@ -10,20 +10,40 @@ export const PACT_FFI_VERSION = '0.3.19';
 let ffi: typeof ffiLib;
 let ffiLogLevel: LogLevel;
 
-const initialiseFfi = (logLevel: LogLevel): typeof ffi => {
+const initialiseFfi = (logLevel: LogLevel, logFile?: string): typeof ffi => {
```
TimothyJones, line 13: I think this would be better if it were more explicit - either you have a log file, or you don't.  The current form means that it's easy to misread the intermediate functions that don't touch log level.

The guideline I would recommend is that optional parameters exist only at a boundary - and their presence (or not) is handled in the function where they are optional.

src/ffi/index.ts
```diff
@@ -10,20 +10,40 @@ export const PACT_FFI_VERSION = '0.3.19';
...
   logger.debug(`Initalising native core at log level '${logLevel}'`);
   ffiLogLevel = logLevel;
-  ffiLib.pactffiInitWithLogLevel(logLevel);
+  if (logFile) {
+    const convertLogLevelToFfiLoglevelFilter = {
```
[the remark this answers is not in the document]
TimothyJones: If this is a constant map, it should:

1) Be declared outside all functions
2) Be all caps

I think you could call it `FFI_LOG_LEVEL_FILTER`. But also, I don't think I understand why it's here. Why not just use `FfiLogLevelFilter` directly?

src/ffi/index.ts
```diff
@@ -10,20 +10,40 @@ export const PACT_FFI_VERSION = '0.3.19';
...
+      warn: FfiLogLevelFilter.LOG_LEVEL_WARN,
+      error: FfiLogLevelFilter.LOG_LEVEL_ERROR,
+    };
+
+    logger.debug(
+      `Setting log file to '${convertLogLevelToFfiLoglevelFilter[logLevel]}'`
+    );
+    const res = ffiLib.pactffiLogToFile(
+      logFile,
+      convertLogLevelToFfiLoglevelFilter[logLevel]
```
TimothyJones, line 30: What happens if `convertLogLevelToFfiLoglevelFilter[logLevel]` is undefined?

TimothyJones: I think this whole inlined `convertLogLevelToFfiLoglevelFilter` constant would be better replaced with a `switch` - it would introduce no new concepts, and would give you error handling for free.

In general, if you need a really long name for something then it's a good signal that there's an alternative design that would be better.

test/consumer.integration.spec.ts
```diff
@@ -188,7 +192,9 @@ describe('FFI integration test for the HTTP Consumer API', () => {
       interaction.withQuery('someParam', 0, 'someValue');
       interaction.withRequestBinaryBody(
         bytes,
-        isWin ? 'application/octet-stream' : 'application/gzip'
```
TimothyJones, line 191: I realise this was already this way when you got here, but there should not be if statements in tests - *especially* on platform.

TimothyJones:
Looks alright to me. I left a number of style comments inline.

> feat: add pactffi_log_to_file for consumer/verifier 

Note that this ends up in the changelog, and this message describes the internals of the change, not how users of pact-js-core will see the change. I think it should describe how the users of pact-js-core will see the change.

> Passed in via an optional param, to getFfiLib which is called by a couple of functions, as an optional param. Wonder if it would be better to have these in an opts object that can take named params.

See my comments inline. I don't think it's a good idea to pass optional parameters down to child functions.

I also think that introducing an options object just for this seems unnecessary (and is kind of a workaround for the child function problem). Is there some reason you can't use the same pattern that the rest of the options uses?

YOU54F:
Thanks for the review dude will address these when I get time :)

test/consumer.integration.spec.ts
YOU54F: Totally accept that and agreed, it points to a wider issue we have with content matching both cross platform and cross arch.

It's important to have these tests in, rather than skipping them, but raising these as discussion points so we can either fix the issue at source in the rust core, or provide platform/arch specific guidance where it is required.

This is a smell here, that is going to affect end users, unless they are aware of it

YOU54F: failing making this change, there are just failing tests for something unrelated to the contributors change, which leads to a sub-optimal experience all round :(

src/ffi/index.ts
YOU54F: bad things probably 😂 or possibly, I don't know. Makes sense to test all the cases. I just did the bare minimum I could to get logging to file into pact-js when I was in the middle of diving into learning about FFI interfaces, so it is most certainly not the correct way, nor finished, I should move it to a draft and then it can be picked up either by me, or others if they feel inclined

native/ffi.cc
YOU54F: agreed 👍🏾

src/ffi/index.ts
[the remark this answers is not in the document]
YOU54F: > But also, I don't think I understand why it's here

I don't know 😅 probably not through an active decision decision, just mushy mushy, get me some loggy

native/ffi.cc
YOU54F: Updated to

```
    case 5:
      logLevel = LevelFilter::LevelFilter_Trace;
      break;
    default:
      std::string err =  "pact-js-core C integration: Unable to parse log level number: ";
      err += number;
      throw Napi::Error::New(env, err);
```

src/ffi/index.ts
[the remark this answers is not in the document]
YOU54F: now using `FfiLogLevelFilter[logLevel]`

test/consumer.integration.spec.ts
YOU54F: all fixed now in the rust core, and resolved in pact-js-core/pact-js (and other consuming libraries)

https://github.com/pact-foundation/pact-reference/pull/429

YOU54F:
> > feat: add pactffi_log_to_file for consumer/verifier
> 
> Note that this ends up in the changelog, and this message describes the internals of the change, not how users of pact-js-core will see the change. I think it should describe how the users of pact-js-core will see the change.

Updates to reflect the end user facing argument that has been added, and to which functions they have been added, which is more sensible 👍🏾 thanks for the comment, totally agree

src/ffi/index.ts
YOU54F: If it is undefined, napi will throw an error as we are expecting a numerical argument

```
Napi::Value PactffiLogToFile(const Napi::CallbackInfo& info) {
   // return: int
   Napi::Env env = info.Env();

  if (info.Length() < 2) {
    throw Napi::Error::New(env, "PactffiLogToFile(envVar) received < 2 arguments");
  }

    if (!info[0].IsString()) {
    throw Napi::Error::New(env, "PactffiLogToFile(envVar) expected a string");
  }

  if (!info[1].IsNumber()) {
    throw Napi::Error::New(env, "PactffiLogToFile(envVar) expected a number");
  }
```

I've also retained the existing `FfiLogLevelFilter[logLevel]` rather than needing to add a new function `convertLogLevelToFfiLoglevelFilter`

src/ffi/index.ts
YOU54F: This is now moved and updated to 

```
export const getFfiLib = (
  logLevel: LogLevel = DEFAULT_LOG_LEVEL,
  logFile: string | undefined = undefined
): typeof ffi => {
```

native/ffi.cc
```diff
@@ -49,6 +49,64 @@ Napi::Value PactffiInitWithLogLevel(const Napi::CallbackInfo& info) {
...
+      break;
+    case 4:
+      logLevel = LevelFilter::LevelFilter_Debug;
+      break;
+    case 5:
+      logLevel = LevelFilter::LevelFilter_Trace;
+      break;
+    default:
+      std::string err =  "pact-js-core C integration: Unable to parse log level number: ";
+      err += number;
```
mefellows, line 76: Is `err` defined somewhere?

src/ffi/index.ts
```diff
@@ -125,18 +121,24 @@ const initialiseFfi = (logLevel: LogLevel): typeof ffi => {
...
+    ffi = initialiseFfi();
+    logger.debug(
+      `Initialising native core at log level '${logLevel}'`,
+      logFile
+    );
+    if (logFile) {
+      logger.debug(
+        `writing log file at level ${FfiLogLevelFilter[logLevel]} to ${logFile}`
       );
-      logger.warn(`The new requested log level '${logLevel}' will be ignored`);
```
mefellows, line 139: Are we missing handling the error in this scenario (somebody initialising the FFI more than once with a different log level)?

mefellows:
Sorry for the late response - just added a couple more comments now, but thanks for the initiative and moving this forward!

PR #432, opened 2023-03-08
fix: Fix an issue where invalid VerifierOptions keys with Array values would cause a TypeError

TimothyJones:
### PR Template

This PR fixes an issue raised by [Gustavo Souza on slack here](https://pact-foundation.slack.com/archives/C9VBGLUM9/p1678279738564239), where an unknown key with an array value would cause the Verifier validation to fail with `cannot read property forEach of undefined`.

It also adds a test for this case.

The test also passes.

### DIFF
--- a/src/verifier/validateOptions.spec.ts
+++ b/src/verifier/validateOptions.spec.ts
@@ -47,6 +47,18 @@ describe('Verifier argument validator', () => {
         });
       });
     });
+
+    context('when given an unknown array argument', () => {
+      it('should return a Verifier object', () => {
+        expectSuccessWith({
+          madeupArg: [''],
+          providerBaseUrl: 'http://localhost',
+          pactBrokerUrl: 'http://foo.com',
+          pactUrls: ['http://idontexist'],
+          provider: 'someprovider',
+        } as VerifierOptions);
+      });
+    });
   });
 
   context('when not given --pact-urls or --provider-base-url', () => {
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -32,12 +32,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - name: Set MSVS version
-        run: npm config set msvs_version 2017
       - name: Use Node.js ${{ matrix.node-version }}
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
+      - name: Set MSVS version
+        run: npm config set msvs_version 2017
       - run: script/ci/build-and-test.sh
         env:
           NODE_VERSION: ${{ matrix.node-version }}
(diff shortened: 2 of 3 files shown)


TimothyJones:
I don't know how the ubuntu build ever worked, but I'll fix that in this PR too.

src/verifier/validateOptions.spec.ts
```diff
@@ -47,6 +47,18 @@ describe('Verifier argument validator', () => {
         });
       });
     });
+
+    context('when given an unknown array argument', () => {
+      it('should return a Verifier object', () => {
+        expectSuccessWith({
```
YOU54F, line 53: Should the user be aware that an invalid option was provided, and bail because of it?

otherwise they might feel that the argument is correctly being set, but it isn't

src/verifier/validateOptions.spec.ts
```diff
@@ -47,6 +47,18 @@ describe('Verifier argument validator', () => {
...
+          madeupArg: [''],
```
YOU54F, line 54: would  js treat `['']` as different to `[]` (because js)

and would that make a diff for this logic `(rules || []).forEach((rule) => {`

src/verifier/validateOptions.spec.ts
TimothyJones: I thought about adding a warning - but the existing behaviour (for unknown options that don't have an array value) was to silently ignore them.

I'm not sure if there are some unvalidated options - if we want to warn, we'll have to double check the list.

src/verifier/validateOptions.spec.ts
TimothyJones: Nope 👍 

`rules` is from the rules object above - which is only accessed using the key. The value is only relevant for entering in the `Array.isArray` branch, which was missing the undefined check.

You can see the same `(rules || [])` pattern on the other branch - I'm not wild about having inline defaults, but I didn't want to change the code too much just to match personal preferences.

src/verifier/validateOptions.spec.ts
YOU54F: hmmm, whilst that wouldn't be my preference, it is the way it is currently so I wouldn't want to break userspace too much. 

looking at the users provided file, the only invalid option I think was `tags` which is now `providerVersionTags` and `log` option is no longer in v10 of pact-js post moving to the rust core.

YOU54F: that is probably one for another PR tho!

mefellows:
Thanks team! 🙏 

We chose not to fail when incorrect arguments were passed because the previous implementation didn't either, and who knows what people would do. Now, in theory, the implementation _did_ fail - just not with a warning, so we could potentially introduce that.

I'm still inclined to leave it as a warn or silently ignore, because there are still plenty of those that would still upgrade from 9.x.x and then run into this behaviour.

Issue #416, opened 2022-11-13
Linux/arm64 CI failing - Cannot find binary for platform linux with architecture arm64


GZLiew:
_Thank you for reporting a bug! We appreciate it very much. Issues are a big input into the priorities for Pact development_

_All italic text in this template is safe to remove before submitting_

_Thanks again!_

### Software versions

_Please provide at least OS and version of pact-js_

- **OS**: _e.g. Linux
- **Node Version**: 16.18.1

### Issue Checklist

Please confirm the following:

- [x] I have upgraded to the latest
- [x] I have the read the FAQs in the Readme
- [x] I have triple checked, that there are **no unhandled promises** in my code
- [x] I have set my log level to debug and attached a log file showing the complete request/response cycle

### Expected behaviour

No errors in CI machine

### Actual behaviour

getting error

Error: Error while locating pact binary: Cannot find binary for platform 'linux' with architecture 'arm64'.
--
  | at throwError (/workdir/node_modules/@pact-foundation/pact-core/standalone/install.ts:7:9)
  | at Object.getBinaryEntry (/workdir/node_modules/@pact-foundation/pact-core/standalone/install.ts:51:9)
  | at Object.standalone (/workdir/node_modules/@pact-foundation/pact-core/src/pact-standalone.ts:40:5)
  | at Object.<anonymous> (/workdir/node_modules/@pact-foundation/pact-core/src/pact-standalone.ts:72:16)
  | at Module._compile (node:internal/modules/cjs/loader:1155:14)
  | at Module._extensions..js (node:internal/modules/cjs/loader:1209:10)
  | at Object.require.extensions.<computed> [as .js] (/workdir/node_modules/ts-node/src/index.ts:1608:43)
  | at Module.load (node:internal/modules/cjs/loader:1033:32)
  | at Function.Module._load (node:internal/modules/cjs/loader:868:12)
  | at Module.require (node:internal/modules/cjs/loader:1057:19)
  | at require (node:internal/modules/cjs/helpers:103:18)
  | at Object.<anonymous> (/workdir/node_modules/@pact-foundation/pact-core/src/server.ts:2:1)
  | at Module._compile (node:internal/modules/cjs/loader:1155:14)
  | at Module._extensions..js (node:internal/modules/cjs/loader:1209:10)
  | at Object.require.extensions.<computed> [as .js] (/workdir/node_modules/ts-node/src/index.ts:1608:43)
[...]

PR #444, opened 2023-05-02
CI: Test across Node 16/18/20 + arm64 MacOS test

YOU54F:
- Consolidates CI test task to use single format for all three OS's
- Updates CI test task to use latest action tags (v3 checkout, v3 node)
- Remove steps no longer needed (updating node-gyp)
- Ensure FFI folder is deleted when running `download-libs.sh` otherwise it asks the user to confirm everytime. 🤯 
- Add a developer note to download the libs, before running `npm ci` when running locally
- Add macOS arm64/m1 runner test via Cirrus-CI
  - Requires rosetta due to using v1.x of pact-ruby-standalone
- Add amd64/x86_64 linux test via Cirrus-CI in preparation for ARM64 linux testing in further PR to be raised
- Test ARM64 Linux, fails as expected as per issue #416

## Smells

Cross CI provider and cross arch has shown a few inconsistencies in content-type matching, which is clearly demonstrable here. Haven't thought of a solution just yet, but this at least highlights and allows us to test more combinations than before

```
const isWin = process.platform === 'win32';
const isLinux = process.platform === 'linux';
const isDarwinArm64 = process.platform === 'darwin' && process.arch === 'arm64';
const isLinuxArm64 = process.platform === 'linux' && process.arch === 'arm64';
const isCirrusCi = process.env['CIRRUS_CI'] === 'true';
const usesOctetStream =
  isLinuxArm64 || isWin || isDarwinArm64 || (isCirrusCi && isLinux);
```
 
 
  See https://github.com/pact-foundation/pact-reference/issues/171 for why we have an OS switch here
  
  ### Previous state
  

>   Windows: does not have magic mime matcher, uses content-type
>   OSX on CI: does not magic mime matcher, uses content-type
>   OSX: has magic mime matcher, sniffs content
>   Linux: has magic mime matcher, sniffs content

 
 ### Current state
  

>   Windows: does not have magic mime matcher, uses content-type
>   OSX arm64: does not magic mime matcher, uses content-type
>   OSX x86_64: has magic mime matcher, sniffs content
>   OSX x86_64 - CI GitHub Actions: has magic mime matcher, sniffs content
>   Linux - CI Cirrus: does not have magic mime matcher, uses content-type

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -50,7 +50,8 @@ mac_task:
       - NODE_VERSION: 16
       - NODE_VERSION: 18
       - NODE_VERSION: 20
-  install_script:
+  install_script: # we need to install rosetta as v1.x of pact-ruby-standalone doesn't support arm64
+    - softwareupdate --install-rosetta --agree-to-license
     - brew install nvm
     - source $(brew --prefix nvm)/nvm.sh
     - nvm install $NODE_VERSION
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -26,7 +26,6 @@ BUILD_TEST_TASK_TEMPLATE: &BUILD_TEST_TASK_TEMPLATE
 #   << : *BUILD_TEST_TASK_TEMPLATE
 
 linux_amd64_task:
-  skip: "changesInclude('.github/**')" 
   env:
     matrix:
       - IMAGE: node:16-slim
@@ -40,7 +39,6 @@ linux_amd64_task:
 
 
 mac_task:
-  skip: "changesInclude('.github/**')" 
   macos_instance:
     image: ghcr.io/cirruslabs/macos-ventura-base:latest
   env:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,58 @@
+env:
+  PACT_BROKER_FEATURES: publish_pacts_using_old_api
+
+BUILD_TEST_TASK_TEMPLATE: &BUILD_TEST_TASK_TEMPLATE
+  arch_check_script:
+    - uname -am
+  test_script:
+    - node --version
+    - script/ci/build-and-test.sh
+
+# These are probably expected to fail in the post install script
+# until we are packing v2.0.0 of pact-ruby-standalone that supports
+# arm64 linux
+linux_arm64_task:
+  skip: "!changesInclude('.github/**')" 
+  env:
+    matrix:
+      - IMAGE: node:16-slim
+      - IMAGE: node:18-slim
+      - IMAGE: node:20-slim
+  arm_container:
+    image: $IMAGE
+  install_script:
+    - apt update --yes && apt install --yes curl python3 make build-essential g++
+  << : *BUILD_TEST_TASK_TEMPLATE
+
+linux_amd64_task:
+  skip: "!changesInclude('.github/**')" 
+  env:
+    matrix:
+      - IMAGE: node:16-slim
+      - IMAGE: node:18-slim
+      - IMAGE: node:20-slim
+  container:
+    image: $IMAGE
+  install_script:
+    - apt update --yes && apt install --yes curl python3 make build-essential g++
+  << : *BUILD_TEST_TASK_TEMPLATE
+
+
+mac_task:
+  skip: "!changesInclude('.github/**')" 
+  macos_instance:
+    image: ghcr.io/cirruslabs/macos-ventura-base:latest
+  env:
+    PACT_BROKER_FEATURES: publish_pacts_using_old_api
+    NVS_HOME: ${HOME}/.nvs
+    PATH: ${NVS_HOME}:${PATH}
+    matrix:
+      - NODE_VERSION: 16
+      - NODE_VERSION: 18
+      - NODE_VERSION: 20
+  install_script:
+    - brew install nvm
+    - source $(brew --prefix nvm)/nvm.sh
+    - nvm install $NODE_VERSION
+    - nvm use $NODE_VERSION
+  << : *BUILD_TEST_TASK_TEMPLATE
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -12,7 +12,7 @@ BUILD_TEST_TASK_TEMPLATE: &BUILD_TEST_TASK_TEMPLATE
 # until we are packing v2.0.0 of pact-ruby-standalone that supports
 # arm64 linux
 linux_arm64_task:
-  skip: "!changesInclude('.github/**')" 
+  skip: "changesInclude('.github/**')" 
   env:
     matrix:
       - IMAGE: node:16-slim
@@ -25,7 +25,7 @@ linux_arm64_task:
   << : *BUILD_TEST_TASK_TEMPLATE
 
 linux_amd64_task:
-  skip: "!changesInclude('.github/**')" 
+  skip: "changesInclude('.github/**')" 
   env:
     matrix:
       - IMAGE: node:16-slim
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -21,7 +21,7 @@ linux_arm64_task:
   arm_container:
     image: $IMAGE
   install_script:
-    - apt update --yes && apt install --yes curl python3 make build-essential g++
+    - apt update --yes && apt install --yes curl python3 make build-essential g++ unzip zip
   << : *BUILD_TEST_TASK_TEMPLATE
 
 linux_amd64_task:
@@ -34,12 +34,12 @@ linux_amd64_task:
   container:
     image: $IMAGE
   install_script:
-    - apt update --yes && apt install --yes curl python3 make build-essential g++
+    - apt update --yes && apt install --yes curl python3 make build-essential g++ unzip zip
   << : *BUILD_TEST_TASK_TEMPLATE
 
 
 mac_task:
-  skip: "!changesInclude('.github/**')" 
+  skip: "changesInclude('.github/**')" 
   macos_instance:
     image: ghcr.io/cirruslabs/macos-ventura-base:latest
   env:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -25,18 +25,18 @@ BUILD_TEST_TASK_TEMPLATE: &BUILD_TEST_TASK_TEMPLATE
 #     - apt update --yes && apt install --yes curl python3 make build-essential g++ unzip zip
 #   << : *BUILD_TEST_TASK_TEMPLATE
 
-# linux_amd64_task:
-#   skip: "changesInclude('.github/**')" 
-#   env:
-#     matrix:
-#       - IMAGE: node:16-slim
-#       - IMAGE: node:18-slim
-#       - IMAGE: node:20-slim
-#   container:
-#     image: $IMAGE
-#   install_script:
-#     - apt update --yes && apt install --yes curl python3 make build-essential g++ unzip zip
-#   << : *BUILD_TEST_TASK_TEMPLATE
+linux_amd64_task:
+  skip: "changesInclude('.github/**')" 
+  env:
+    matrix:
+      - IMAGE: node:16-slim
+      - IMAGE: node:18-slim
+      - IMAGE: node:20-slim
+  container:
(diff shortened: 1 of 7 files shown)


.cirrus.yml
```diff
@@ -0,0 +1,60 @@
+env:
```
TimothyJones, line 1: I guess this is here to do builds on the systems that github doesn't currently support?

Whatever the reason is, using two CI systems is worth documenting, probably in comments in both files. Otherwise a future developer might make changes to one CI script but not the other.

.github/workflows/build-and-test.yml
```diff
@@ -8,64 +8,30 @@ on:
 
 jobs:
   build-and-test-osx:
-    runs-on: macos-latest
+    runs-on: ${{ matrix.os }}
```
TimothyJones, line 11: AWESOME!! I had been meaning to do this for ages, but never got around to figuring out what the syntax was.

.github/workflows/build-and-test.yml
```diff
@@ -8,64 +8,30 @@ on:
...
-          npm prefix -g | % {npm config set node_gyp "$_\node_modules\node-gyp\bin\node-gyp.js"}
-        shell: pwsh
-      - run: bash script/ci/build-and-test.sh
-        shell: bash
-        env:
-          NODE_VERSION: ${{ matrix.node-version }}
-          # The windows build agent has trouble unpacking the tar for
-          # linux, so we only download windows binaries. This means
-          # we cannot release from Windows in CI.
           ONLY_DOWNLOAD_PACT_FOR_WINDOWS: true
```
TimothyJones, line 37: I don't know if `ONLY_DOWNLOAD_PACT_FOR_WINDOWS` is still  used, probably it can be removed.

.github/workflows/build-and-test.yml
```diff
@@ -8,64 +8,30 @@ on:
...
-
-    steps:
-      - uses: actions/checkout@v2
-      - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v1
-        with:
-          node-version: ${{ matrix.node-version }}
-      - name: Set MSVS version
-        run: npm config set msvs_version 2017
       - run: script/ci/build-and-test.sh
```
TimothyJones, line 32: This needs to be `bash script/ci/build-and-test.sh`, otherwise it won't work on windows.

.github/workflows/build-and-test.yml
```diff
@@ -8,64 +8,30 @@ on:
...
+        if: runner.os == 'Windows'
         env:
```
TimothyJones, line 34: I think you also need 

```
        shell: bash
```

for windows

DEVELOPER.md
```diff
@@ -3,6 +3,12 @@
...
 ```
+bash script/download-libs.sh
 npm ci
 npm test
-```
\ No newline at end of file
+```
+
+
+_notes_ - As a developer you need you download the FFI libraries prior to running `npm install` / `npm ci` as the libraries will be expected to be there, and you won't have any `node_modules` installed yet.
```
TimothyJones, line 12: I think this paragraph would be improved by saying that `script/download-libs.sh` does this download - otherwise someone might think that it was a separate instruction here.

script/lib/download-ffi.sh
```diff
@@ -17,7 +17,7 @@ if [[ $(find "${FFI_DIR}" -name "${FFI_VERSION}*") ]]; then
 fi
 
 warn "Cleaning ffi directory $FFI_DIR"
-rm -rf "${FFI_DIR:?}/*"
```
TimothyJones, line 20: There's a reason I did this, but I don't remember what it was. It might be that I didn't want to remove the folder in some cases. As long as it works this way, it's fine.

Also, now that I re-read this line it could be improved with an error message:

```
rm -rf "${FFI_DIR:?FFI_DIR was not set}"
```

Not super important, as the `:?` is only there as a safeguard if `set -u` is removed or didn't take for some reason.

TimothyJones: Update: Actually, this script should start:

```
#!/bin/bash -eu
set -e
set -u
```

because the github windows bash doesn't respect the shebang line.

test/message.integration.spec.ts
```diff
@@ -14,8 +14,12 @@ chai.use(chaiAsPromised);
...
 
 const isWin = process.platform === 'win32';
-const isOSX = process.platform === 'darwin';
-const isCI = process.env['CI'] === 'true';
+const isLinux = process.platform === 'linux';
+const isDarwinArm64 = process.platform === 'darwin' && process.arch === 'arm64';
+const isLinuxArm64 = process.platform === 'linux' && process.arch === 'arm64';
+const isCirrusCi = process.env['CIRRUS_CI'] === 'true';
+const usesOctetStream =
+  isLinuxArm64 || isWin || isDarwinArm64 || (isCirrusCi && isLinux);
```
TimothyJones, line 22: I remain uncomfortable with doing the test conditionally - but I don't really see what can be done to improve it. Probably needs a fix in the Rust core so that this concern doesn't leak to consumers of the FFI.

TimothyJones (approved):
Nice work! Left a bunch of comments + suggestions.

.github/workflows/build-and-test.yml
YOU54F: you can set it at a defaults level at the top level of a job, so it applies to all run commands 👍🏾 saves adding bash to every step because windows

script/lib/download-ffi.sh
YOU54F: That was a gem of a tim's tech tid bit I found, as my windows scripts were unusually unflakey. they can't all pass first time right 😅 yep some errorz were being swallowed.

I checked to see if there was anything else in the folder we needed that might get trashed but it seemed not, I know deleting the whole standalone folder is a no no as it contains the `install.ts` script

test/message.integration.spec.ts
YOU54F: The alternative is skip the test completely.

Having the test has value, and the fact that is shows discrepancy between os's/arches means we have a fighting chance of fixing it, there is a tracking issue of the rust core, I'll use this as a show case / test bed

YOU54F: the other alternative is leave it as is, and contributors get a sub optimal experience on some arches, as they can't run the repo, and end users get a shitty experience because we don't solve the root issue, which is either a rust core either or just an operating system / arch oddity and all users of pact need guidance on it. 

It doesn't feel like, and it smells, but that pong is enough to get it cleaned up. 

I can only shave so many yaks in a day tho buddy

YOU54F: > Probably needs a fix in the Rust core so that this concern doesn't leak to consumers of the FFI.

If it is possible to fix in the core, it's like docker for mac and windows not being able to use localhost to talk to each other

https://docs.docker.com/desktop/networking/#use-cases-and-workarounds-for-all-platforms

maybe it's just one of those things we don't wish to be, but just _is_

DEVELOPER.md
YOU54F: shall add, feel free to propose something nice, I just quickly boshed something in whilst it was fresh in my mind, i'll make it more verbose tomorrow and add some decent docs about cirrus-ci, notes about keeping them in sync (gh actions / cirrus ci) and tips for running both locally (via act or cirrus-cli) - depending on the end users machine 👍🏾 

We can massively improve contributions if we improve on the docs on helping people pick up the codebase and just get rocking.

.github/workflows/build-and-test.yml
YOU54F: I didn't scan, but will check 👍🏾

test/message.integration.spec.ts
TimothyJones: Yeah. This one is more of a comment than something you can do anything about.

.github/workflows/build-and-test.yml
YOU54F: I have been making some mental foo of late, you can generate matrices dynamically in your actions, so I have a run of pact ffi/ pact verifier cli and pact mock server cli, for all the targets, in one super clean action.

I just wish you could report on individual matrix combos, because you could build up an awesome compat matrix, with github badges. 

The other way is to have separate workflows , I need to check if you can pull in one master workflow and then have loads of mini workflows, that just specify x combo (node 14 / windows) so they appear as seperate workflows, and you could get a status badge for each one.

DEVELOPER.md
TimothyJones: I would just change:

> As a developer you need you download the FFI libraries

to

> As a developer, you need to run `script/download-libs.sh` to download the FFI libraries...

.cirrus.yml
YOU54F: correcto, it is currently only open source and free to use m1 runners, they also provide arm64 linux runners (as well as freebsd), and coupled with the ability to run the workflows locally in ephemeral VM's, it makes for a pleasant experience

Circle CI offers arm64 linux, but only x86_64 macs, so cirrus is simple free way we can get arm64 runners for linux and macos, and also allow contributors to perform the same runs, on forks, with cirrus too.

If we got AWS runners or something like that, it would exclude users who don't have the ability to pay for that kind of compute.

.github/workflows/build-and-test.yml
TimothyJones: Hm ok. Last time I looked at this, Windows needed both:

```
run: bash script/ci/build-and-test.sh
shell: bash
```

but perhaps it has improved.

YOU54F:
> Nice work! Left a bunch of comments + suggestions.

appreciate the 👀 buddy ❤️

.github/workflows/build-and-test.yml
YOU54F: Ahh I've only ever needed `shell: bash` - I use `bash` prefix when I use powershell or command prompt in windows, although it tends to get annoyed and I have to do `bash -c 'annoying folder with spaces'`

because `Program Files` I mean seriously dudes

.github/workflows/build-and-test.yml
YOU54F: It is for unpacking the standalones (windows fails to unpack the linux tar.gz) and it used in the tests to only test for the windows binary, if that env var is set

.cirrus.yml
YOU54F: Will add some developer docs to cover this platform in an upcoming pr :)

test/message.integration.spec.ts
```diff
@@ -14,8 +14,12 @@ chai.use(chaiAsPromised);
 const { expect } = chai;
 
 const isWin = process.platform === 'win32';
-const isOSX = process.platform === 'darwin';
-const isCI = process.env['CI'] === 'true';
+const isLinux = process.platform === 'linux';
+const isDarwinArm64 = process.platform === 'darwin' && process.arch === 'arm64';
+const isLinuxArm64 = process.platform === 'linux' && process.arch === 'arm64';
+const isCirrusCi = process.env['CIRRUS_CI'] === 'true';
+const usesOctetStream =
```
mefellows, line 21: This makes me sad. Not your fault of course. We need to find a solution to this!

YOU54F: I know, I'll card this up for a yak shave, now have we have a way of testing repeatably across platform, we can sort it, or document it

.github/workflows/build-and-test.yml
TimothyJones: > I just wish you could report on individual matrix combos, because you could build up an awesome compat matrix, with github badges.

I dunno - I think it's better to just report the whole success / failure - as then you don't get tempted to continue with "well, it doesn't work on everything we deploy to".

It would be nice to be able to generate the list of tested combinations, though!

YOU54F: a whole just fail isn’t helpful for me finding the needle in the haystack and github’s view of multi matrix runs needs some work, detail gets lots and you just have 12 boxes that al look the same from the truncated description until you click on one, and that is an expensive use of time

i need to check out github blocks again as there is meant to be some cool stuff you can do in readmes now

PR #450, opened 2023-06-14
chore!: Remove pactflow binary, as it causes confusion between the open source Pact and the PactFlow product, which are not the same thing.

TimothyJones:
I removed the PactFlow binary, as it probably shouldn't have been there in the first place.

I think if you (PactFlow) need this, it should be a separate npm package published by PactFlow.

Also, there's no documentation on it that I can see.


YOU54F:
> I removed the PactFlow binary, as it probably shouldn't have been there in the first place.

it was added as it was included in the pact-ruby-standalone, which in bundled, so the command was exposed

> I think if you (PactFlow) need this, it should be a separate npm package published by PactFlow.

why should it be? it a tool to publish pacts to a broker, either. pact broker or pactflow broker. there may be other brokers in the future, should be exclude all others, bar the OG pact broker?

The code included in open source, although it publishes to a closed source hosted platform, the code is visible on github, and freely able to be viewed and amended by users.

open source gives users the freedom, this takes away the users choice to upload their pacts to any supported broker (open source or otherwise), the code included here, is all MIT.

> Also, there's no documentation on it that I can see.

There is documentation in the source repo -

we link to it here 

https://github.com/pact-foundation/pact-js-core#cli-tools

it was an omission not updating the readme to include it in the list of tools

It is very clear in the source documentation linked to the bundler,  which points to the included gems, pact-broker_client contains the pactflow code

https://github.com/pact-foundation/pact_broker-client#pact-broker-client

>A client for the [Pact Broker](https://docs.pact.io/pact_broker/) and [PactFlow](https://pactflow.io/?utm_source=ossdocs&utm_campaign=pact_broker_client_readme). Publishes and retrieves pacts, pacticipants, pacticipant versions, environments, deployments and releases. Supports publishing provider contracts for PactFlow.

>All commands prefixed with pact-broker can be used with the OSS Pact Broker and PactFlow. Commands prefixed with pactflow can only be used with PactFlow.

Not sure if your suggestion is we should duplicate documentation, or point to source.

I had no issue with PactFlow being involved in Pact open source projects prior to working for PactFlow, as it didn't take anything away from the open source offering. I equally have no issue with it now, but want to ensure that our work in PactFlow causes benefit to Pact.

For example some of the work to improve the next steps of the pactflow command output, benefits Pact Broker users as they now have clearer colours to denote information and coloured urls for next steps. I know it is only a little thing but every little helps.

YOU54F:
>This is technically a breaking change, so I marked it as such. I don't know if these binaries are actually used - they're not in the Pact docs.

What do you class as the Pact docs?

https://docs.pact.io/pact_broker/client_cli/readme#provider-contracts-pactflow-only

source https://github.com/pact-foundation/pact_broker-client#provider-contracts-pactflow-only

YOU54F:
> Remove pactflow binary, as it causes confusion between the open source Pact and the PactFlow product

Who is confused by the binary?

TimothyJones:
I have made the case for this clearly and repeatedly- mostly elsewhere, but let me do it again here:

If pactflow have interest in maintaining good relationships with unpaid volunteers, pactflow should be very interested in making a clear distinction between the open source project and the pactflow projects.

Surprise links to pactflow (either in code or in documentation) erode this distinction. 

This should never have been committed in the first place.

mefellows:
Thanks all. As discussed in the recent strategy sessions, this distinction has been tabled as one of the key priorities to address. The initial "sprinkling" of PactFlow in various places initially may have been unwelcome by some, so let's approach this more deliberately in response. For clarity, the proposal is that we do an audit of all PactFlow/Pact touch points, collaborate and agree on a communications and integration plan going forward that works for all parties.

PR #453, opened 2023-07-06
Fix/improve native logging lookup

YOU54F:
this avoids the user having an error for the path lookups, when traversing each of the available binding paths, unless all paths are exhausted.

this way, if the library is available and working fine, users don't see an error in the logs.

if for whatever reason the library isn't loaded, an error message in displayed with some useful tips

### DIFF
--- a/src/ffi/index.ts
+++ b/src/ffi/index.ts
@@ -60,7 +60,7 @@ const bindingPaths = [
 let ffiLib: Ffi;
 
 const renderBinaryErrorMessage = (error: unknown) => {
-  logger.debug(`${supportedPlatformsMessage}`);
+  logger.debug(supportedPlatformsMessage);
   logger.error(`Failed to find native module for ${platform}: ${error}`);
   bindingPaths.forEach((bindingPath) => {
     logger.debug(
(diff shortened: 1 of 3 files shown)


src/verifier/nativeVerifier.ts
```diff
@@ -11,10 +11,10 @@ import { setupVerification } from './argumentMapper';
 const pkg = require('../../package.json');
 
 export const verify = (opts: VerifierOptions): Promise<string> => {
-  const ffi = getFfiLib(opts.logLevel);
   if (opts.logLevel) {
     setLogLevel(opts.logLevel);
   }
+  const ffi = getFfiLib(opts.logLevel);
```
mefellows, line 17: that'll do it 😆

mefellows (approved):

src/ffi/index.ts
```diff
@@ -92,25 +91,28 @@ const initialiseFfi = (logLevel: LogLevel): typeof ffi => {
...
+          logger.info(
+            'pact native library successfully found, and the correct version',
+            ffiLib.pactffiVersion()
+          );
+          return false;
         }
+        return true;
       } catch (error) {
-        renderBinaryErrorMessage(error, bindingPath);
+        // renderBinaryErrorMessage(error, bindingPath);
```
mefellows, line 107: Will merge and then remove this comment.

PR #454, opened 2023-07-07
fix: drop support for node 15 and earlier

mefellows:
This might actually be best being a `fix!` and a new major version. I'm actually unclear on the appropriate semver bump required for this.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -21,10 +21,10 @@ PREBUILD_AND_TEST: &PREBUILD_AND_TEST
   test_16_script:  NODE_VERSION=16 script/ci/build-and-test.sh && script/ci/clean.sh
 
 linux_arm_task:
+  only_if: $CIRRUS_CHANGE_TITLE !=~ 'chore\(release\):.*'
   env:
     SET_NVM: "true"
   arm_container:
-  # container:
     image: node:20-slim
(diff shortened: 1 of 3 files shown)


YOU54F:
>This might actually be best being a fix! and a new major version. I'm actually unclear on the appropriate semver bump required for this.

@TimothyJones has suggested exactly that, breaking change and major semver bump (can't find the exact comments but they are around)

Relates to https://github.com/pact-foundation/pact-js/issues/1028

Have updated. also removed the erroneous comment as mentioned in the last PR and updated cirrus ci, to not run after the release has been committed back to git, this matches up with github

YOU54F (approved):
Thanks Matt, have updated to force a major version bump, with `fix!` so will get this released now.

TimothyJones:
Breaking change, so yes, needs a major bump. Nice work.

I think it’s a good general practice to follow the Bodens release schedule

mefellows:
> Breaking change, so yes, needs a major bump. Nice work.

Agree in general. I think where it was a grey area for me, is that we have dropped support for previous versions but never officially via the `engines` flag. So it would have failed, just in a way that wasn't a first class "unsupported" failure. 

But I think it's good to be explicit. We'll need to major bump Pact JS if this gets pulled in then also.

PR #491, opened 2024-02-23
feat: v2 pact-standalone with arm64 linux/macos support (pact-node branch)

### LINKED ISSUE #489: RFC: Update pact-node branch with latest ruby standalone binaries and support multiple architectures
## Previous issues

- https://github.com/pact-foundation/pact-js-core/issues/487
- https://github.com/pact-foundation/.github/issues/17

## Problem

Some users are unable to migrate to the latest version of pact-js/pact-js-core which leverages the Rust runtime, due to various reasons.

They are beginning to encounter errors with the standalone package as it is outdated, and predates our work on the traveling-ruby fork to support the latest ruby runtimes and multiple archs.

##  Steps

- PR to update `pact-js-core` to consume v2 `pact-ruby-standalone`
- PR to update `pact-js-core` to support multiple arches (arm64 for linux and macos)
- Release of pact-node from `pact-node` branch
- Release of `pact-js` v9 to include new release of `pact-node`

##  Breaking Changes

None

### User Actions

None

##  Potential Downsides/Caveats

Supporting the pact-node branch and pact-js v9.x.x even for an interim release, would probably warrant a dependency update.

We would probably not want to provide tier 1 support for this, but it may help unblock users as an interim measure

## Has it previously been tested

Yes, when I created the new traveling-ruby releases, I tested against the pact-node branch in my fork.

https://github.com/YOU54F/pact-js-core/commits/pact-node-arm64/

Released to npm 

https://www.npmjs.com/package/@you54f/pact-node?activeTab=code

and consumed in a pact-js-core fork 

https://www.npmjs.com/package/@you54f/pact/v/9.20.0?activeTab=code

## Other options

Users can provide their own binary [docs ref](https://github.com/pact-foundation/pact-js-core/tree/pact-node?tab=readme-ov-file#pact-download-location) and rename the latest release to match whatever is in pact-node's latest release.

However it is checksum validated. Another potential could be to remove the checks here, such that users can provide their own binaries, and we wouldn't have to do updates of pact-node / pact-js-core v9, if the standalone is updated.

> For the example given above, Pact will look for the binary at /home/some-user/Downloads/pact-1.44.0-win32.zip for a Windows system. However, by using this method, you must use the correct Pact version binary associated with this version of Pact-Node. For extra security measurements, checksum validation has been added to prevent tampering with the binaries.

YOU54F:
fixes #489


mefellows:
Awesome, thanks! I've manually run the release locally - so it should now be available 🎉

PR #494, opened 2024-03-13
fix: 🐛 avoid node-gyp rebuild install script

danymarques:
Add an "install" script to "override" npm behavior of adding a "node-gyp rebuild" install script on the fly when publishing.


.DS_Store
YOU54F, line 1: could you remove the .DS_Store files and i can give this a merge

YOU54F (changes requested):

YOU54F:
thanks for the change dude and the investigation. it would be nice to add in the additional context to the pr for posterity

YOU54F (approved):

Issue #447, opened 2023-05-12
Binary Content Matching is odd - cross platform/arch


YOU54F:
## Smells

Cross CI provider and cross arch has shown a few inconsistencies in content-type matching, which is clearly demonstrable here. Haven't thought of a solution just yet, but this at least highlights and allows us to test more combinations than before

Encountered in https://github.com/pact-foundation/pact-js-core/pull/444#issue-1692947149

If you have code that looks like this, there is a definite code-smell

https://github.com/pact-foundation/pact-js-core/blob/facfef810036d0b80fb3c2fbcafd8efa255b6171/test/consumer.integration.spec.ts#L23-L34

The condition is used in these tests

https://github.com/pact-foundation/pact-js-core/blob/dac0e46eafc5cdf633f459adc591e9cc5ba26c24/test/consumer.integration.spec.ts#L77-L87

https://github.com/pact-foundation/pact-js-core/blob/dac0e46eafc5cdf633f459adc591e9cc5ba26c24/test/message.integration.spec.ts#L94-L109

>  See https://github.com/pact-foundation/pact-reference/issues/171 for why we have an OS switch here
    Windows: does not have magic mime matcher, uses content-type
    OSX on CI: does not magic mime matcher, uses content-type
    OSX: has magic mime matcher, sniffs content
     Linux: has magic mime matcher, sniffs content
 
 So based on our compat table of 
 
 | OS                     | Amd64 | Arm64 |
|--------------- |---------|-----|
| Linux (glibc)    | ✅          | ✅   |
| Mac                  | ✅          | ✅  |
| Windows          | ✅          | ❌   | 
| Linux (alpine)  | ❌          | ❌ | 

✅ = supported
❌ = not supported

We have a few different ways we can test to cover different arch combinations

- Locally
- Locally with GitHub Actions / Act
- GitHub Actions
- Locally with Cirrus CLI
- Cirrus CI
     
| OS                     | Amd64 | Arm64 |   Amd64 (GH) | Amd64 (GH Act) | Amd64 (Cirrus) | Arm64 (Cirrus CI) |    Amd64 (Cirrus CLI) | Arm64 (Cirrus CLI) |  
|--------------- |---------|-----|-----|-----|-----|-----|-----|-----|
| Linux (glibc)    | ❓         | ❓   |❓   |❓   |❓   |❓   |❓   |❓   |
| Mac                  | ❓          | ❓  |❓   |❌   |❓   |❓   |❓   |❓   |
[...]

PR #507, opened 2024-05-13
feat: alpine support

### LINKED ISSUE #490: RFC: Support Alpine
## Previous issues

- https://github.com/pact-foundation/pact-js/issues/1032
- https://github.com/pact-foundation/pact-js/issues/1026
- https://github.com/pact-foundation/pact-js/issues/899#issuecomment-1462945915

## Problem

Alpine users are unable to utilise the `pact-js-core` and `pact-js`, since we switched to a Rust backed core powered by [Pact FFI](https://github.com/pact-foundation/pact-reference/tree/master/rust/pact_ffi)

This causes inconvenience to both users, and maintainers, and friction in general for Pact adoption. 

It would be nice to provide at least preliminary support for Alpine with our Pact FFI.

The Pact FFI library, since 

- 0.4.16 has published both x86_64 and aarch64 musl variants of the Pact FFI libary in static form `.a`
- 0.4.17 has published converted `.so` musl variants

##  Steps

- Create Alpine based node prebuilds
  - We utilise `prebuildify` and native bindings, to be able to interact with our Pact FFI shared library. We pre-create these bindings in CI, so that our end-users do not need to build themselves (which requires Python/GCC and others)
- Test Alpine based images in CI
  - We should test official Alpine node images, for the node versions we support

##  Breaking Changes

- Package size will be increased by ~90mb
  - ~45mb per musl arch (x86_64/aarch64)

### User Actions

None

### Support levels

We would probably consider this less supported than our existing glibc based versions, therefore we would probably want to ask that users recreate an issues encountered in Alpine, 

- Reproducible 
  - Dockerfiles for any issue encountered, so a maintainer can reproduce 
  - Add test to repo test suite
  - Add github repo, with CI, showing the issue
- Alpine specific
  - Create the same example on a glibc based distro.

The following would allow us to determine if it is a general Pact issue, or one specific to Alpine. 

I am unsure of how much we will be able to affect, that is Alpine specific, however we can compile a list of caveats/workarounds or bugs that appear out of them.

### Pact Plugins

The Pact plugin ecosystem will not work properly on musl based systems until the following PR's are merged

- https://github.com/pact-foundation/pact-plugins/pull/56
- https://github.com/pact-foundation/pact-plugins/pull/57
- https://github.com/pactflow/pact-protobuf-plugin/pull/47

Without the above PR's, the pact-plugin drivers will attempt to download non musl variants of the plugins, which are unlikely to work on Alpine without glibc/gcompat.

After the PR's are merged, the plugin-cli will attempt to download a musl variant, and fallback to the vanilla linux variant instead.

### Package sizes

It may be the case that package sizes are too large, when installed on a target machine, as we store all the prebuilds in the published package.

There are a few options we could take

- Strip binaries
  -  _may_ cause issues
- Don't package pre-builds in pact-js-core published package
  - Allow prebuildify to fetch required pre-build from GitHub releases
  - Allow user to specify path to their own prebuild (currently possible with PACT_PREBUILD_LOCATION)

I know we used to download the standalone on the fly, back in older versions of pact-js/pact-node but this would cause issues in many corporate environments with proxies.

## Has it previously been tested

Yes, but with modified musl variants of the plugin framework as per my above PR's

- https://github.com/YOU54F/pact-js-core/tree/alpine
- https://github.com/YOU54F/pact-js-core/actions/runs/7880984396/job/21503869458

The plugin tests are currently skipped on this repo, but were enabled in my testing, to try and test it out end to end.

There were a few node based workarounds depending on the version used, and there is still an issue with some of the plugin tests on some variant combos against node 16.  

## Other options

- We do nothing
  - Users are unable to utilise Pact on Alpine, or have to use compatibility based workarounds which can result in weird behaviour
  - Users have to switch from using Alpine, to a glibc based distro. For some users/organisations, this isn't always easy, especially when trying to integrate Pact into an established ecosystem.

YOU54F:
fixes #490 

- Download musl variants of libpact_ffi
- Provide a `node.napi.musl.node` for `aarch64` & `x86_64` musl variants
- Prepackaged in pact-js-core, so end users don't need to do anything to enable, will autodetect if running under musl or glibc
- Content-type matching (related to #447) resolved by https://github.com/pact-foundation/pact-reference/pull/429
- Re-enabled all plugin tests in this repo
  - Plugin tests passing on alpine


mefellows:
Amazing - will try and look today Yousaf! 👏

YOU54F:
No rush at all dude!

YOU54F:
switching this to a draft, as there are a couple of additional things I want to do

- switch to musl `*.so` (smaller)

but few changes I want to make in pact-ref with regards to some linker errors and content-type issues

https://github.com/pact-foundation/pact-reference/issues/416#issuecomment-2113121731

https://github.com/YOU54F/pact-reference/pull/3

src/ffi/index.ts
```diff
@@ -4,7 +4,7 @@ import logger, { DEFAULT_LOG_LEVEL } from '../logger';
 import { LogLevel } from '../logger/types';
 import { Ffi } from './types';
 
-export const PACT_FFI_VERSION = '0.4.20';
+export const PACT_FFI_VERSION = '0.4.26';
```
YOU54F, line 7: update to lib_pactffi 0.4.21 on release

script/lib/download-ffi.sh
```diff
@@ -8,7 +8,7 @@ require_binary gunzip
 
 require_env_var FFI_VERSION
 
-BASEURL=https://github.com/pact-foundation/pact-reference/releases/download
+BASEURL=https://github.com/you54f/pact-reference/releases/download
```
YOU54F, line 11: revert back to pact-foundation on release of lib_pactffi 0.4.21.

Cannot use current version of 0.4.20 due to https://github.com/pact-foundation/pact-reference/issues/436

script/download-plugins.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash -eu
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd)" # Figure out where the script is running
+
+. "${SCRIPT_DIR}/lib/export-binary-versions.sh"
+"${SCRIPT_DIR}/install-plugin-cli.sh"
+$HOME/.pact/bin/pact-plugin-cli install -y https://github.com/mefellows/pact-matt-plugin/releases/tag/$PACT_PLUGIN_MATT_VERSION
```
YOU54F, line 6: should we add matt plugin the pact plugin index to save having to install it manually here and in the pact-go test suite?

.github/workflows/build-and-test.yml
YOU54F, line 1: builds and tests amd64/arm64 linux versions of glibc + musl variants of vanilla node images (-slim / -alpine respectively)

binding.gyp
```diff
@@ -1,150 +1,185 @@
...
-        "<(module_root_dir)/ffi",
-      ],
-      "conditions": [
-        [
-          "OS=='win'",
-          {
-            "libraries": [
-              "<(module_root_dir)/ffi/pact_ffi.dll.lib"
+    "variables": {
+        "is_alpine": "<!(grep -q Alpine /etc/os-release && echo true || echo false)"
```
YOU54F, line 3: This file is essentially the same despite the white space changes.

We check if the os is alpine and then build the musl bindings.

node-gyp-build detects musl at runtime when loading the libraries, so will correctly pick between musl/glibc as required

script/ci/prebuild-alpine.sh
```diff
@@ -0,0 +1,35 @@
...
+PREBUILD_NAME="node.napi"
+
+apk add bash curl python3 make g++
+
+. "${SCRIPT_DIR}/../lib/export-binary-versions.sh"
+"${SCRIPT_DIR}/../lib/download-ffi.sh"
+rm -rf build node_modules
+npm ci --ignore-scripts
+export npm_config_target=${NODE_VERSION}
+# ar -s ffi/*/libpact_ffi.a
```
YOU54F, line 17: we can remove this commented line and the one below it

script/ci/build-and-test.sh
```diff
@@ -19,7 +19,6 @@ fi
 
 node --version
 npm --version
-
```
YOU54F, line 22: remove

.gitignore
```diff
@@ -56,6 +56,7 @@ ts-node-*
 
 # FFI native bindings
 *.so
+*.a
```
YOU54F, line 59: remove

src/ffi/internals/index.spec.ts
YOU54F, line 1: these aren't used

src/ffi/internals/index.ts
YOU54F, line 1: these aren't used

test/consumer.integration.spec.ts
```diff
@@ -82,9 +75,10 @@ describe('FFI integration test for the HTTP Consumer API', () => {
         .request({
           baseURL: `http://${HOST}:${port}`,
           headers: {
-            'content-type': usesOctetStream
-              ? 'application/octet-stream'
-              : 'application/gzip',
+            // 'content-type': usesOctetStream
```
YOU54F, line 78: remove

test/consumer.integration.spec.ts
```diff
@@ -225,9 +220,10 @@ describe('FFI integration test for the HTTP Consumer API', () => {
         .request({
           baseURL: `http://${HOST}:${port}`,
           headers: {
-            'content-type': usesOctetStream
-              ? 'application/octet-stream'
-              : 'application/gzip',
+            'content-type': 'application/gzip',
```
YOU54F, line 223: remove

test/consumer.integration.spec.ts
```diff
@@ -202,7 +196,8 @@ describe('FFI integration test for the HTTP Consumer API', () => {
       interaction.withQuery('someParam', 0, 'someValue');
       interaction.withRequestBinaryBody(
         bytes,
-        usesOctetStream ? 'application/octet-stream' : 'application/gzip'
+        'application/gzip'
+        // usesOctetStream ? 'application/octet-stream' : 'application/gzip'
```
YOU54F, line 200: remove

test/consumer.integration.spec.ts
```diff
@@ -225,9 +220,10 @@ describe('FFI integration test for the HTTP Consumer API', () => {
         .request({
           baseURL: `http://${HOST}:${port}`,
           headers: {
-            'content-type': usesOctetStream
-              ? 'application/octet-stream'
-              : 'application/gzip',
+            'content-type': 'application/gzip',
+            // 'content-type': usesOctetStream
```
YOU54F, line 224: remove

test/verifier.integration.spec.ts
```diff
@@ -175,7 +175,8 @@ describe('Verifier Integration Spec', () => {
     // thread '<unnamed>' panicked at 'Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context
     // with RUST_BACKTRACE=1 it seems that it relates to fetching from the broker, and something bad
     // is happening in reqwest
-    context.skip('from a Pact Broker', () => {
+    const skipPluginTests = process.env['SKIP_PLUGIN_TESTS'] === 'true';
```
YOU54F, line 178: this is still flakey, haven't worked out what is causing it

PR #510, opened 2024-05-28
fix: set process.stdout._handle.setBlocking(true) on ffi init

YOU54F:
fixes https://github.com/pact-foundation/pact-js/issues/1216

Large number of pact verifications fail when run in parallel


mefellows:
The only additional comment I would make is, do we need to consider undoing that operation when pact shuts down? Probably unnecessary.

YOU54F:
I've assumed it's killed with the node process for the sake of simplicity of getting a fix out. Certainly happy to go for a belt and braces approach (being set at a test lifecycle start and end, rather than when the ffi init is called )

PR #604, opened 2025-01-20
ci: use arm runners for arm64 workflows

YOU54F:
arm64 runners now available, so lets use them to speed up our CI runs

https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/


mefellows:
Thank you sir!

PR #648, opened 2025-06-11
chore(deps): update everything

JP-Ellis:
Update all the dependencies in one go, to help get through the backlog.

Other than the upgrades, the only notable change is removing the `airbnb` eslint configs as they do not support ESLint 9.


YOU54F:
the upgrade of `eslint` excludes  node 16, which is eol anyway, we can bump but would need to be a breaking change and note in the `engines` section of the package.json

`rimraf` looks like it has an issue on windows

`mocha` looks it like has an issue with node 18, this went eol 3 months ago, so we could lock the node engines to 20.

It seems a shame as these are test dependencies and having pact-js _more_ compatible with older applications would be preferable for me or needing to run the latest great dev deps, especially when we rarely rely on new features, so it would only be bug fixes, and as pact-js-core is used as a testing tool and we wouldn't expect it to be used in production, I am less worried

hmmm 🤔

JP-Ellis:
Thanks for taking a look! Sorry I started trying to fix this but didn't see it through.

YOU54F:
> Thanks for taking a look! Sorry I started trying to fix this but didn't see it through.

no need to apologise at all buddy!

sometimes when I’m hoovering up, i get distracted and start cleaning something else 😅

PR #861, opened 2026-05-01
feat: Update FFI lib to 0.5.4 and add the addInteractionReference function

rholshausen:
Adds the new `addInteractionReference` function which provides the capability to record external references against a Pact interaction. This allows linking the interactions to external references like OpenAPI or AsyncAPI.


mefellows:
The build is passing, but it's because the JS interface isn't executed so it won't expose that the native library isn't available to the JS runtime.

You'll need to add it into the relevant C++ files also:

* https://github.com/pact-foundation/pact-js-core/blob/master/native/consumer.h
* https://github.com/pact-foundation/pact-js-core/blob/master/native/consumer.cc

and export it to the JS interface here:

* https://github.com/pact-foundation/pact-js-core/blob/master/native/addon.cc

I should update the DEVELOPER.md docs to create some better pointers around this (pun not intended).

PR #885, opened 2026-05-25
chore!: drop Node 18/20 support, target Node 22+

JP-Ellis:
Bring `pact-js-core` in line with `pact-js` by:

- Replacing eslint+prettier with biome for both formatting and linting
- Replacing mocha+chai with vitest for testing
- Dropping support for unmaintained versions of Node. We now target Node 22, 24 and 26 as of the time of this PR's creation


JP-Ellis:
@mefellows when reviewing, I would _highly_ recommend you look at the individual commits. This will help see the pattern, given this has a large number of automated changes.

src/ffi/node-gyp-build.ts
```diff
@@ -1 +0,0 @@
-declare module 'node-gyp-build';
```
mefellows, line 1: I don't quite recall why this was needed, hopefully it won't affect the ffi build.

mefellows:
Nice one Josh!

PR #931, opened 2026-07-31
fix(deps): pact-ffi 0.5.6 (rebuild)

YOU54F:
Bumps `pact-ffi` to 0.5.6, and wires up FFI functions that had accumulated
unexposed across several releases.

## Header review

Diffing the generated `pact.h` between `libpact_ffi-v0.5.4` and `v0.5.5`: no
functions were removed and no existing signature changed. The additions are all
from upstream's plugin observability work (#537):

- `pactffi_set_test_run_id`
- `pactffi_register_plugin_log_callback`
- `pactffi_get_plugin_logs`

Every `pactffi_*` symbol referenced from `native/` still exists in 0.5.5, so the
bump itself is backwards compatible.

## Newly exposed

Auditing all 280 exported functions against what `native/` actually binds turned
up a handful that were usable but never surfaced, including some missed when
earlier bumps landed:

| Function | Available since | Exposed as |
| --- | --- | --- |
| `pactffi_set_key` | pre-0.5 | `setKey` |
| `pactffi_using_plugin_with_delay` | 0.5.0 | `addPluginWithDelay` |
| `pactffi_verifier_set_follow_redirects` | 0.5.3 | `followRedirects` verifier option |
| `pactffi_get_tls_ca_certificate` | ≤0.5.4 | `getTlsCaCertificate` |
| `pactffi_set_test_run_id` | 0.5.5 | `setTestRunId` |

`addPluginWithDelay` is the interesting one for flaky tests: plugins are not
necessarily ready to accept connections the moment `addPlugin` returns, and this
gives callers a way to wait for that startup work.

`getTlsCaCertificate` was already declared in `src/ffi/types.ts` but had no
binding behind it, so calling it would have thrown. It now has one.

The remaining unexposed functions are deliberate: the pact *reading* API (which
this library has no use for as a pact writer), and deprecated variants
superseded by `_v2` forms that are already wired up.

## Not exposed, and why

Two of the three functions added in 0.5.5 are unusable as shipped:

- `pactffi_register_plugin_log_callback` takes `Option<PluginLogCallback>`, which
  cbindgen emits as an *opaque* struct passed by value. A caller cannot construct
  a value of an incomplete type, so the function cannot be called at all.
  Reported as pact-foundation/pact-reference#544.
- `pactffi_get_plugin_logs` needs a `plugin_instance_id`, and nothing in the
  header returns one — the only place it is ever surfaced is the first argument
  of the callback above, so #544 gates it.
  Reported as pact-foundation/pact-reference#545.

## ⚠️ TLS mock servers regress in 0.5.5

Upstream `bdfe6f68` is described as fixing "`pactffi_create_mock_server_for_transport`
is unable to start a TLS mock server". In practice it makes things worse, and
this is worth knowing before treating 0.5.5 as a TLS fix:

| version | `createMockServer(host, port, tls: true)` |
| --- | --- |
| 0.5.4 | returns a valid port |
| 0.5.5 | panics in rustls, returns `-4` (`CORE_PANIC`) |

The panic is `Could not automatically determine the process-level CryptoProvider
from Rustls crate features`. It also poisons an internal mutex, so a *subsequent*
FFI call on the same pact handle hits `panic in a function that cannot unwind`
and aborts the host process.

Reported as pact-foundation/pact-reference#546.

A TLS integration test is included but marked `describe.skip`, with a comment
explaining why — left enabled it kills the test worker rather than failing.

## Testing

`setKey` has an integration test asserting the key reaches the V4 pact file, and
`followRedirects` has validation tests. Locally the suite is 88 passed / 1 failed,
the failure being a pre-existing multipart test that fails identically on 0.5.4
and is unrelated to this change.

### DIFF
--- a/native/addon.cc
+++ b/native/addon.cc
@@ -15,6 +15,7 @@ Napi::Object Init(Napi::Env env, Napi::Object exports) {
   exports.Set(Napi::String::New(env, "pactffiMockServerMismatches"), Napi::Function::New(env, PactffiMockServerMismatches));
   exports.Set(Napi::String::New(env, "pactffiCreateMockServerForTransport"), Napi::Function::New(env, PactffiCreateMockServerForTransport));
   exports.Set(Napi::String::New(env, "pactffiCleanupMockServer"), Napi::Function::New(env, PactffiCleanupMockServer));
+  exports.Set(Napi::String::New(env, "pactffiGetTlsCaCertificate"), Napi::Function::New(env, PactffiGetTlsCaCertificate));
   exports.Set(Napi::String::New(env, "pactffiWritePactFile"), Napi::Function::New(env, PactffiWritePactFile));
   exports.Set(Napi::String::New(env, "pactffiWritePactFileByPort"), Napi::Function::New(env, PactffiWritePactFileByPort));
   exports.Set(Napi::String::New(env, "pactffiNewPact"), Napi::Function::New(env, PactffiNewPact));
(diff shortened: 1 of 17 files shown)


YOU54F:
I assume this needs a rebase to pull in #937

mefellows:
looks like a legit failure, but only seems to be affecting Ubuntu 🤔

YOU54F:
tls issue now fixed, will get this merged and released now :)

